PR #2, opened 2024-07-05
formatutils: ensure we dont leak name ptr

gulafaran:
store format name in a std::string and free the char ptr to ensure we dont leak on each call to drmGetFormatName.


vaxerski (approved):
my bad, forgot it's a newly allocated string.

PR #3, opened 2024-07-06
add missing virtual destructor to base class IAllocator

gulafaran:
add missing virtual destructor to base class IAllocator and also add destructor to CGBMAllocator and free gbm_device


vaxerski:
you added a commit from another mr #2 here

gulafaran:
> you added a commit from another mr #2 here

opsie, fixed.

vaxerski (approved):
tanks!

PR #7, opened 2024-07-07
Add  scheduleFrame reasons

UjinT34:
Adds an optional `scheduleFrameReason` argument to `scheduleFrame`. Now it is used for tracing only. Later it can be used for stuff like `cursor:no_break_fs_vrr`


vaxerski (approved):
sure

PR #8, opened 2024-07-08
CMakeLists.txt: require native hyprwayland-scanner

MarkusVolk:
This would fix a cross-compile issue where hyprwayland-scanner is pulled in for target but needs to run on host.


vaxerski (approved):
tanks

PR #9, opened 2024-07-08
Fix cursor swapchain buffer creation

UjinT34:
Use `attrs` instead of `params` since `guessFormatFrom` makes changes to `attrs` and other code also uses this var.
Fixes cursor buffer creation on nvidia.
Fixes `[ERR] [EGL] Command eglQueryDmaBufModifiersEXT errored out with EGL_BAD_PARAMETER (0x12300): EGL_BAD_PARAMETER error: In eglQueryDmaBufModifiersEXT: Invalid format` on hl side but rendering to this buffer still fails.


vaxerski (approved):
thanks, true oops

PR #10, opened 2024-07-10
pass explicitOutFence pointer and do not touch the value

UjinT34:
Handle `explicit*Fence` as wlroots does. `setExplicitOutFence` doesn't need an argument but I kept it for a while. It should be called when there is some active `wait_timeline`... whatever this might be.

### DIFF
--- a/src/backend/drm/impl/Atomic.cpp
+++ b/src/backend/drm/impl/Atomic.cpp
@@ -101,8 +101,8 @@ void Aquamarine::CDRMAtomicRequest::addConnector(Hyprutils::Memory::CSharedPoint
     add(connector->crtc->id, connector->crtc->props.active, enable);
 
     if (enable) {
-        if (connector->output->supportsExplicit && STATE.explicitOutFence >= 0)
-            add(connector->crtc->id, connector->crtc->props.out_fence_ptr, STATE.explicitOutFence);
+        if (connector->output->supportsExplicit && STATE.committed & COutputState::AQ_OUTPUT_STATE_EXPLICIT_OUT_FENCE)
+            add(connector->crtc->id, connector->crtc->props.out_fence_ptr, (uintptr_t)&STATE.explicitOutFence);
 
         if (connector->crtc->props.gamma_lut && data.atomic.gammad)
             add(connector->crtc->id, connector->crtc->props.gamma_lut, data.atomic.gammaLut);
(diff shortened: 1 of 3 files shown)


src/backend/drm/impl/Atomic.cpp
```diff
@@ -101,8 +101,8 @@ void Aquamarine::CDRMAtomicRequest::addConnector(Hyprutils::Memory::CSharedPoint
     add(connector->crtc->id, connector->crtc->props.active, enable);
 
     if (enable) {
-        if (connector->output->supportsExplicit && STATE.explicitOutFence >= 0)
-            add(connector->crtc->id, connector->crtc->props.out_fence_ptr, STATE.explicitOutFence);
+        if (connector->output->supportsExplicit && STATE.committed & COutputState::AQ_OUTPUT_STATE_EXPLICIT_OUT_FENCE)
+            add(connector->crtc->id, connector->crtc->props.out_fence_ptr, (uintptr_t)&STATE.explicitOutFence);
```
vaxerski, line 105: why are you committing the out fence if you just commented out the setter? maybe not comment the setter out but this instead?

UjinT34: It commits the pointer to a value so it can be written by something external. At least I understand it this way. That's what wlroots is doing. The setter only signals that this pointer is needed when some timeline is used.

vaxerski: OH, alright.

vaxerski: Make it not cleared on ::commit then maybe in COutputState? that clears it

UjinT34: wlroots clears it in `drm_connector_state_init`. I guess it is relevant across one frame only. If this `drm_connector_state_init` is called every frame. I am confused %)

vaxerski: IIRC it is. You can clear it in the CDRMOutput::commit just before impl->commit() then

UjinT34: wlroots set the timeline and uses it  in `drm_atomic_connector_prepare` to get `primary_in_fence_fd`. aquamarine gets `explicitInFence` directly from hl. Clearing it before impl->commit() would be too early. I think it is fine to leave it as is.

vaxerski: but now hyprland can't access out because it's cleared instantly after connector->commitState on L1379
image

state->onCommit clears the outFence

UjinT34: hmm... looks like with the current api hl should do the clearing

vaxerski (approved):
alright, thanks, though now needs a hl mr doesnt it

PR #12, opened 2024-07-11
Expose bo buffer mapping

UjinT34:
Expose gbm buffer mapping through `beginDataPtr`. The mapping is created only once so only flags passed for the first time are used. Can be used to copy data to the gbm buffer if rendering fails (`flags = GBM_BO_TRANSFER_WRITE`). Assumes 4 bytes per pixel as the most common case for cursor buffers.

### DIFF
--- a/src/allocator/GBM.cpp
+++ b/src/allocator/GBM.cpp
@@ -194,10 +194,20 @@ SDMABUFAttrs Aquamarine::CGBMBuffer::dmabuf() {
 
 std::tuple<uint8_t*, uint32_t, size_t> Aquamarine::CGBMBuffer::beginDataPtr(uint32_t flags) {
     uint32_t dst_stride = 0;
-    if (!boBuffer)
+    if (boBuffer)
+        allocator->backend->log(AQ_LOG_ERROR, "beginDataPtr is called a second time without calling endDataPtr first. Returning old mapping");
+    else
         boBuffer = gbm_bo_map(bo, 0, 0, attrs.size.x, attrs.size.y, flags, &dst_stride, &gboMapping);
     // FIXME: assumes a 32-bit pixel format
     return {(uint8_t*)boBuffer, attrs.format, attrs.size.x * attrs.size.y * 4};
+}
+
+void Aquamarine::CGBMBuffer::endDataPtr() {
+    if (gboMapping) {
+        gbm_bo_unmap(bo, gboMapping);
+        gboMapping = nullptr;
+        boBuffer   = nullptr;
+    }
 }
 
 CGBMAllocator::~CGBMAllocator() {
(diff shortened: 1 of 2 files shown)


vaxerski:
Why? It's really slow. I'd rather provide an API to submit CPU buffers for cursors.

UjinT34:
A temporary solution. Not for the hw cursors themselves but for features that require them.
`beginDataPtr` api already exists. Is there any better way of returning something useful from this function?

src/allocator/GBM.cpp
```diff
@@ -188,6 +192,14 @@ SDMABUFAttrs Aquamarine::CGBMBuffer::dmabuf() {
     return attrs;
 }
 
+std::tuple<uint8_t*, uint32_t, size_t> Aquamarine::CGBMBuffer::beginDataPtr(uint32_t flags) {
+    uint32_t dst_stride = 0;
+    if (!boBuffer)
+        boBuffer = gbm_bo_map(bo, 0, 0, attrs.size.x, attrs.size.y, flags, &dst_stride, &gboMapping);
+    // FIXME: assumes cursor with DRM_FORMAT_ARGB8888
```
vaxerski, line 199: comment is wrong: assumes buffer with a 32-bit pixel format

src/allocator/GBM.cpp
```diff
@@ -188,6 +192,14 @@ SDMABUFAttrs Aquamarine::CGBMBuffer::dmabuf() {
     return attrs;
 }
 
+std::tuple<uint8_t*, uint32_t, size_t> Aquamarine::CGBMBuffer::beginDataPtr(uint32_t flags) {
+    uint32_t dst_stride = 0;
+    if (!boBuffer)
+        boBuffer = gbm_bo_map(bo, 0, 0, attrs.size.x, attrs.size.y, flags, &dst_stride, &gboMapping);
```
vaxerski, line 198: if we call map twice, does that leak?

UjinT34: It doesn't leak and doesn't update the flags. I've mentioned that. It creates the mapping on the first call and then just returns it every time. Currently there is no need in changing the flags.

vaxerski: yes then question 2 applies, does it update when we change the buffer with the gpu?

UjinT34: With `GBM_BO_TRANSFER_READ` or `GBM_BO_TRANSFER_READ_WRITE` it should. But I didn't test it.

vaxerski: why not just unmap in `endDataPtr`?

vaxerski (approved):
lgtm, thanks!

PR #14, opened 2024-07-15
handle monitor hotplugs and fix a few bugs that it revealed

ikalco:
realized this while working on hyprwm/Hyprland#6875

### DIFF
--- a/src/backend/Session.cpp
+++ b/src/backend/Session.cpp
@@ -303,46 +303,47 @@ void Aquamarine::CSession::dispatchUdevEvents() {
         return;
     }
 
+    dev_t              deviceNum = udev_device_get_devnum(device);
+    SP<CSessionDevice> sessionDevice;
+    for (auto& sDev : sessionDevices) {
+        if (sDev->dev == deviceNum)
+            sessionDevice = sDev;
+    }
+
+    if (!sessionDevice) {
+        udev_device_unref(device);
+        return;
+    }
+
     if (action == std::string{"add"})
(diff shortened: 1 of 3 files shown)


nonetrix:
Is this related to comment I left where monitor would go black after turning it on and off unless I moved a window or anything happened on it?

src/backend/Session.cpp
```diff
@@ -303,46 +303,47 @@ void Aquamarine::CSession::dispatchUdevEvents() {
         return;
     }
 
+    dev_t              deviceNum = udev_device_get_devnum(device);
+    SP<CSessionDevice> sessionDevice;
+    for (auto& sDev : sessionDevices) {
+        if (sDev->dev == deviceNum)
+            sessionDevice = sDev;
```
vaxerski, line 310: why is there no break here?

src/backend/drm/DRM.cpp
```diff
@@ -1439,9 +1448,13 @@ bool Aquamarine::CDRMOutput::setCursor(SP<IBuffer> buffer, const Vector2D& hotsp
         return false;
     }
 
-    if (!buffer)
+    if (!buffer) {
+        connector->crtc->pendingCursor.reset();
+        connector->crtc->cursor->front.reset();
```
vaxerski, line 1453: we can't yank a DRM buffer from KMS while it's being used, will cause glitches.

vaxerski (approved):
thanks!

PR #21, opened 2024-07-20
Remove redundant needsFrame & allow frame schedule skipping on mouse move

UjinT34:
- Removes `needsFrame = true` before `scheduleFrame` since it is set inside. Useful for logging `needsFrame` changes.
- Adds optional `skipShedule` argument to `moveCursor` to allow `cursor:no_break_fs_vrr` with hw cursors

`skipShedule` requires HL updates to be useful. Just passing `true` to `moveCursor` is not enough to fix VRR. Probably needs something like https://github.com/hyprwm/Hyprland/pull/6877 for hw cursors.

### DIFF
--- a/include/aquamarine/backend/DRM.hpp
+++ b/include/aquamarine/backend/DRM.hpp
@@ -194,7 +194,7 @@ namespace Aquamarine {
         virtual bool                                                      test();
         virtual Hyprutils::Memory::CSharedPointer<IBackendImplementation> getBackend();
         virtual bool                                                      setCursor(Hyprutils::Memory::CSharedPointer<IBuffer> buffer, const Hyprutils::Math::Vector2D& hotspot);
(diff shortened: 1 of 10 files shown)


vaxerski (commented):
I fail to see a point. Why not just... not move the cursor?

UjinT34:
It is not guaranteed that the cursor position will be updated to the correct value when something finally decides that the frame should be rendered. Something needs to update the cursor position between frames without triggering extra frames. I think it's easier to skip frame schedule than to call an extra `moveCursor` under complex conditions.
Not moving the cursor will definitely solve the VRR issues but that's not what humans want. They can't keep their hands off the mouse and tend to get disappointed when their cursor is stuck in place %)

vaxerski:
ok

vaxerski (approved):

PR #29, opened 2024-07-24
Fix gbm cursor for nvidia

UjinT34:
Make sure that format is supported by cursor plane even if it is not renderable.
Otherwise rendering will almost freeze if hardware cursors are enabled with nvidia.


src/allocator/GBM.cpp
```diff
@@ -123,6 +126,12 @@ Aquamarine::CGBMBuffer::CGBMBuffer(const SAllocatorBufferParams& params, Hypruti
         }
         bo = gbm_bo_create_with_modifiers2(allocator->gbmDevice, attrs.size.x, attrs.size.y, attrs.format, explicitModifiers.data(), explicitModifiers.size(), flags);
 
+        if (!bo && CURSOR) {
+            // allow non-renderable cursor buffer
```
vaxerski, line 130: would be nice to clarify it's because nvidia

vaxerski (approved):

PR #30, opened 2024-07-25
gbm: manually set modifier for implicit bo allocation

moetayuko:
Ported from wlroots

Fixes https://github.com/hyprwm/Hyprland/issues/7001

GBM_BO_USE_CURSOR fails gbm_bo_create for me so I dropped it.

### DIFF
--- a/src/allocator/GBM.cpp
+++ b/src/allocator/GBM.cpp
@@ -53,8 +53,7 @@ static SDRMFormat guessFormatFrom(std::vector<SDRMFormat> formats, bool cursor) 
 }
 
 Aquamarine::CGBMBuffer::CGBMBuffer(const SAllocatorBufferParams& params, Hyprutils::Memory::CWeakPointer<CGBMAllocator> allocator_,
-                                   Hyprutils::Memory::CSharedPointer<CSwapchain> swapchain) :
-    allocator(allocator_) {
+                                   Hyprutils::Memory::CSharedPointer<CSwapchain> swapchain) : allocator(allocator_) {
     if (!allocator)
         return;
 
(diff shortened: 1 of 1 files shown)


src/allocator/GBM.cpp
```diff
@@ -159,7 +163,7 @@ Aquamarine::CGBMBuffer::CGBMBuffer(const SAllocatorBufferParams& params, Hypruti
     }
 
     attrs.planes   = gbm_bo_get_plane_count(bo);
-    attrs.modifier = gbm_bo_get_modifier(bo);
+    attrs.modifier = hasModifier ? gbm_bo_get_modifier(bo) : fallbackModifier;
```
vaxerski, line 166: why? It should return implicit or linear.

src/allocator/GBM.cpp
```diff
@@ -149,7 +148,12 @@ Aquamarine::CGBMBuffer::CGBMBuffer(const SAllocatorBufferParams& params, Hypruti
 
         if (!bo) {
             allocator->backend->log(AQ_LOG_ERROR, "GBM: Allocating with modifiers failed, falling back to implicit");
```
vaxerski, line 150: with the change, the log is wrong. More like `modifier-less allocation`

src/allocator/GBM.cpp
moetayuko: gbm_bo_get_modifier returns DRM_FORMAT_MOD_INVALID for gbm_bo_create allocated ones as shown in my previous log https://github.com/hyprwm/Hyprland/issues/7001#issuecomment-2247752003
(yea gbm_bo_create succeeded)
```
[hc] loadThemeStyle: loading for size 48
[hc] getShapesC: found 1 images for left_ptr
[LOG] [AQ] GBM: Automatically selected format AR24 for new GBM buffer
[LOG] [AQ] GBM: Buffer is marked as multigpu, forcing linear
[ERR] [AQ] GBM: Allocating with modifiers failed, falling back to implicit
[LOG] [AQ] GBM: Allocated a new buffer with size [Vector2D: x: 128, y: 128] and format AR24 with modifier 72057594037927935 aka INVALID
```

the above hunk adds GBM_BO_USE_LINEAR to flags, but gbm_bo_get_modifier still returns INVALID (I have experimented locally) so explicitly setting it to LINEAR is necessary or the cursor is still invisible. Maybe some code elsewhere expects the modifier not to be INVALID

vaxerski: ok. Makes sense.

moetayuko:
Found a similar patch here https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/1151

vaxerski (approved):
thanks!

PR #32, opened 2024-07-25
fix destroying headless outputs

ikalco:
fixes hyprwm/Hyprland#6995

Hyprland still holds ref to output so it isn't destroyed by std::erase, and destroy event isnt sent


vaxerski (commented):
should also be done for wayland :)

vaxerski:
also dont remove the emit in the dtor, no problem with emitting twice

ikalco:
wayland output destroy already had it but not in dtor so added that

vaxerski (approved):
thanks!

PR #38, opened 2024-08-03
Update switch state in libinput event

toamz:
This fixes a bug, where the aquamarine switch state is never updated, thus the check `if (ENABLED == hlDevice->switchy->state)` blocks all "disabled" events (the default state) and only switch "enabled" events are fired.


vaxerski (approved):
thanks, lol, my bad

PR #50, opened 2024-08-17
buffer align size to stride

gulafaran:
calculate the size to the stride we get from gb_bo_map to better align it.


vaxerski (commented):
comment above return can be removed

gulafaran:
removed

vaxerski (approved):
thanks tom

PR #51, opened 2024-08-17
drm: clear buffer on creation for scanout DRM buffers

vaxerski:
Fixes https://github.com/hyprwm/Hyprland/issues/6950 but I havent tested this at all so pls do


UnixTeletype:
Tested, working fine no white flash thanks a lot my eyes are safe now!! <3

PR #58, opened 2024-08-20
output avoid crashing on no crtc

gulafaran:
dual gpus can init with no crtc and later rescan for them, however getRenderFormats might be called before that happends and null ptr deref, guard against it.


vaxerski (approved):
tanks

PR #61, opened 2024-08-21
add missing destructor to base class IDRMImplementation

gulafaran:
add missing destructor to base class IDRMImplementation


gulafaran:
il refer to the comment in https://github.com/hyprwm/Hyprland/pull/7451#issuecomment-2302883688 hence why gcc warns about it.

PR #68, opened 2024-08-29
don't connect before crtcs are guaranteed

ikalco:
### better "workflow"

scan connectors to get them
assign crtcs if they don't have them
then connect


vaxerski (approved):
sure

PR #72, opened 2024-09-01
allow disabling commit, even if pageFlipPending

ikalco:
fixes hyprwm/Hyprland#7599

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -1465,7 +1465,7 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
                 backend->backend->log(AQ_LOG_DEBUG, std::format("drm: Disabling output {}", name));
         }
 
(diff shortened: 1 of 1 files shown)


src/backend/drm/DRM.cpp
```diff
@@ -1465,7 +1465,7 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
                 backend->backend->log(AQ_LOG_DEBUG, std::format("drm: Disabling output {}", name));
         }
 
-        if ((NEEDS_RECONFIG || (COMMITTED & COutputState::eOutputStateProperties::AQ_OUTPUT_STATE_BUFFER)) && connector->isPageFlipPending) {
+        if (!STATE.enabled && (NEEDS_RECONFIG || (COMMITTED & COutputState::eOutputStateProperties::AQ_OUTPUT_STATE_BUFFER)) && connector->isPageFlipPending) {
```
vaxerski, line 1468: didnt you mean the opposite? This will now err out later in drm if it's enabled and we commit too fast

ikalco: somehow it works both ways but your logic;s right

vaxerski (approved):

PR #87, opened 2024-09-24
output: add a default mode for headless outputs

PaideiaDilemma:
Fixes https://github.com/hyprwm/Hyprland/issues/7276
Fixes https://github.com/hyprwm/hyprlock/issues/434

I tested explicit modes for headless outputs and they are still created properly.

Should this mode set as preferred? It currently is.


vaxerski (approved):
sure :)

PR #95, opened 2024-10-11
Avoid excessive atomic properties updates

UjinT34:
Splitted `addConnector` into 3 parts. Modesetting properties and cursor properties go into separate functions. Mixing those in without need might cause increased cpu usage,  flickering or break vrr. Most noticeable with intel videocards.
Moved `Colorspace` and `hdr_output_metadata` into modeset category.
Removed unnecessary `Colorspace` and `hdr_output_metadata` clearing unless explicitly requested.

Submitting any cursor plane property will break vrr on nvidia. This PR skips setting those properties unless something changes. Doesn't fully fix vrr with hw cursors on nvidia but makes it better. Looks like it is a driver bug and nothing else can be done on AQ/HL side. If this will result in brightness fluctuations with vrr and hw cursors enabled than before this PR it was broken and monitor was stuck at max refresh rate. Should be fixed by disabling vrr to have the same result as before the PR or by disabling hw cursors to get fully working vrr without brightness flickers.


vaxerski:
> results in a new surface commit event

_what_? We're in DRM here, a "surface" doesn't exist. What are you talking about?

If you don't commit the cursor plane state, I assume it's left unaltered...? Have you checked that?

UjinT34:
"planeProps cursor -> listener_commitWindow -> damageSurface -> CDRMAtomicImpl::commit -> planeProps cursor" goes in a loop. I don't know why it works this way. `planeProps(connector->crtc->cursor, ...)` make HL think that a new frame should be rendered and rendering a "new" frame results in a new `planeProps(connector->crtc->cursor, ...)`. Something similar happens with `sendPointerFrame` and locked cursors https://github.com/hyprwm/Hyprland/pull/6877

Looks like it's left unaltered. At least the cursor stays visible and there are no artifacts present.

vaxerski:
> planeProps cursor -> listener_commitWindow

excuse me?

UjinT34:
You send cursor position and get a `_CWlSurfaceCommit` which ends up in a `listener_commitWindow`. No idea why this happens and how to stop it from happening. At least it looks this way. Hard to debug it since something happens outside HL and AQ and triggers this unwanted `_CWlSurfaceCommit`.

vaxerski:
yea like thats the question how the fuck is that related, idk if curing the symptom will help with fixing the underlying issue

UjinT34:
Looks like every or most atomic properties keep their values unless altered so there is no need to repeat them on every commit.
Extra surface commits aren't observed anymore. But mouse movement still breaks vrr on nvidia.
Changes to `Colorspace` and `hdr_output_metadata` should fix increased cpu usage on intel (also requires HL update)

include/aquamarine/output/Output.hpp
```diff
@@ -55,6 +55,9 @@ namespace Aquamarine {
             AQ_OUTPUT_STATE_CTM                = (1 << 10),
             AQ_OUTPUT_STATE_HDR                = (1 << 11),
             AQ_OUTPUT_STATE_DEGAMMA_LUT        = (1 << 12),
+            AQ_OUTPUT_STATE_WCG                = (1 << 13),
```
vaxerski, line 58: wtf is wcg? It's not used as far as I can see

src/backend/drm/impl/Atomic.cpp
```diff
@@ -376,8 +422,18 @@ bool Aquamarine::CDRMAtomicImpl::commit(Hyprutils::Memory::CSharedPointer<SDRMCo
 
     if (ok) {
         request.apply(data);
-        if (!data.test && data.mainFB && connector->output->state->state().enabled && (flags & DRM_MODE_PAGE_FLIP_EVENT))
+        if (!data.test && data.mainFB && connector->output->state->state().enabled && (flags & DRM_MODE_PAGE_FLIP_EVENT)) {
             connector->isPageFlipPending = true;
+            // if (connector->output->state->state().committed & COutputState::AQ_OUTPUT_STATE_CURSOR_POS) {
```
vaxerski, line 427: why this

vaxerski (commented):
generally more or less fine

include/aquamarine/output/Output.hpp
UjinT34: wide color gamut

include/aquamarine/output/Output.hpp
```diff
@@ -78,6 +81,7 @@ namespace Aquamarine {
         };
 
         const SInternalState& state();
+        bool                  cursorCommited = false;
```
vaxerski, line 84: is there any point for this to be publicly r/w?
can this just be a `bool cursorCommitted()`?

Reason is less potential ABI breaks

vaxerski:
rest is ok to me

vaxerski (approved):
thanks!

PR #96, opened 2024-10-12
DRM/Output/Core: Add HW plane infrastructure

vaxerski:
This is required for a better fix to nvidia hw cursors, and generally enables powerful optimizations in the future.

Very wip, not done yet


vaxerski:
I also need to rework how renderer's gpu scheduling is done on hyprland that would be helpful for those "further optimizations" but oh well.

vaxerski:
also should explore drm_dumb buffers first if those don't offer acceptable performance levels, we shouldn't waste gp planes

PR #105, opened 2024-11-11
Fix fences according to spec

UjinT34:
* `setExplicitOutFence` is never called and doesn't really set the fence. Renamed it to reflect the real purpose
* changed fence type to int32_t

> prop_out_fence_ptr
> 
>  Sync File fd pointer representing the outgoing fences for a CRTC. Userspace should provide a pointer to a value of type s32, and then cast that pointer to u64.
https://www.kernel.org/doc/html/v6.11/gpu/drm-kms.html
https://drmdb.emersion.fr/properties/3435973836/OUT_FENCE_PTR
https://drmdb.emersion.fr/properties/4008636142/IN_FENCE_FD


vaxerski (approved):
abi break

Issue #107, opened 2024-11-13
NULL pointer dereference in atomic DRM backend


ziyao233:
Hyprland 0.45.0 together with aquamarine 0.4.4, hyprutils 0.2.4, segfaults on start. The binary is built by clang 19.1.3 and linked against libc++. Hyprland was started with

```
HYPRLAND_NO_CRASHREPORTER=1 HYPRLAND_TRACE=1 AQ_TRACE=1 Hyprland
```

and corresponding log is attached[1]. Some utf8-unclean lines are stripped and they're unrelated to this problem.

Examining the core, it's found that an invalid pointer is dereferenced in `Aquamarine::CDRMAtomicRequest::commit()`

```
$ lldb Hyprland --core core
(lldb) target create "Hyprland" --core "core"
Core file '/home/ziyao/packaging/change_aquamarine/core' (x86_64) was loaded.
(lldb) bt
* thread #1, name = 'Hyprland', stop reason = signal SIGSEGV: address not mapped to object
  * frame #0: 0x00007f217f8e426a libaquamarine.so.3`Aquamarine::CDRMAtomicRequest::commit(unsigned int) + 74
    frame #1: 0x00007f217f8e4f34 libaquamarine.so.3`Aquamarine::CDRMAtomicImpl::reset() + 596
    frame #2: 0x00007f217f8c4dd8 libaquamarine.so.3`Aquamarine::CDRMBackend::start() + 24
    frame #3: 0x00007f217f8797b9 libaquamarine.so.3`Aquamarine::CBackend::start() + 217
    frame #4: 0x000055896fce6e1b Hyprland`CCompositor::initServer(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, int) + 1407
    frame #5: 0x000055896fddfeda Hyprland`main + 1707
(lldb)
```

More digging shows it is `this->conn` dereferenced in `src/backend/drm/impl/Atomic.cpp`[2] that contains a NULL address, leading to a segmentation fault,

```C++
    if (auto ret = drmModeAtomicCommit(backend->gpu->fd, req, flagssss, &conn->pendingPageFlip); ret) {
        backend->log((flagssss & DRM_MODE_ATOMIC_TEST_ONLY) ? AQ_LOG_DEBUG : AQ_LOG_ERROR,
                     std::format("atomic drm request: failed to commit: {}, flags: {}", strerror(-ret), flagsToStr(flagssss)));
        return false;
    }
```

[...]

PR #108, opened 2024-11-18
drm: Validate conn before dereference in CDRMAtomicRequest::commit()

ziyao233:
During startup, CDRMAtomicImpl::reset() may emit a call to method commit of a CDRMAtomicRequest instance with member "conn" uninitialized, leading to a segfault. Validate the the pointer before dereference it as a workaround.

Fixes: 55ac962 ("DRM: preliminary atomic support")
Closes: https://github.com/hyprwm/aquamarine/issues/107


vaxerski (approved):
thanks!

PR #114, opened 2024-12-09
allow multigpu blit from explicit to implicit

ikalco:
fixes hyprwm/Hyprland#8354

this does
* errors before alloc when plane or backend doesn't support requested format in gbm
* allow implicit modifier buffers to be allocated by not forcing linear in gbm
* allow implicit modifiers to be added to glFormats
* allow the blit to_buffer to use external formats (cause implicit is external)


vaxerski (approved):
sure.

PR #106, opened 2024-11-13
scan only cards and not outputs, safeguard against null renderer

gulafaran:
seems pointless scanning the card[0-9]-eDP etc outputs for kms/bootvga/rendernodes just use the [0-9] cards, also safegaurd in commitstate for missing renderer prevents a coredump if __EGL_VENDOR_LIBRARY_FILENAMES is set to mesa on a laptop with dual gpus and a monitor attached to the nvidia dgpu. it might be a way to make this handled better but now it doesnt segfault atleast with a null dereference to renderer.

PR #121, opened 2024-12-18
drm: moved null check for renderer to shouldBlit() (#109)

### LINKED ISSUE #109: Hyprland fails to start in a QEMU VM Hyprland 0.45+ aquamarine 5.0  -- black screen no errors
Updating Arch (and NIXOS) to Hyrprland 0.45.* and aquamarine  0.4.5+ resulted in black screen with no errors in a Proxmox QEMU VM.  

Was forced to roll back to older versions to restore function 
image

The recent DRM changes seem to be the only ones that might relate to this 
https://github.com/hyprwm/aquamarine/commit/3b00e96f90cb0040de6d88ad99bf5f4d443f0c59

https://github.com/hyprwm/aquamarine/commit/545f506a6331e4a35ec16561c10509bf43bb9328

I tried various combination of hyprland and aquamarine but only rolling back to aquamarine 0.4.3  and hyprland 0.45.0  For NIXOS I had to go back to 0.44.  

Anything newer just results in back screen 
image

Normally that screen appears for a few seconds then hyprland loads. 

This is using SDDM for arch, selecting hyprland not hyprland-systemd (which has never worked for me)   In NIXOS I use greetd which also yields the same result. 

I am using Proxmox but others using Virt Manger w/QEMU are reporting the same issues

richen604:
closes #109

early return added in #106 broke `virtio-vga-gl`, moves the check to blitting

tested locally with nixos both via qemu and host configuration with no issues

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -1419,11 +1419,6 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
         return false;
     }
 
-    if (!backend->rendererState.renderer) {
-        backend->backend->log(AQ_LOG_ERROR, "drm: No renderer attached to backend");
-        return false;
-    }
-
     const auto&    STATE     = state->state();
     const uint32_t COMMITTED = STATE.committed;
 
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
right, we don't need the renderer otherwise.

PR #122, opened 2024-12-21
only fail INVALID format when it's set as that

### LINKED ISSUE #117: Disabled interal display doesn't turn black
_Please correct me if this the wrong place to file this issue_

I have two outputs an internal laptop display (connected at eDP-1) and an external monitor (connected at DP-1). I want to disable the internal laptop display and keep the external monitor active as my primary output. Before aquamarine 0.5.1 the following lines in hyprland.conf worked correctly to achieve this:
```
# disable internal display (eDP-1)
monitor=desc:BOE 0x0C8E,disable
# run remaining monitors (e.g. connected through DP-1) with the highest refresh rate and without GSYNC
monitor=,highrr,auto,1,vrr,0
```

Since version 0.5.1 the internal display keeps showing TTY output and the external monitor displays Hyprland. The internal display went black (as i wanted it) on earlier versions of aquamarine (e.g. version 0.5.0). When i switch to another virtual terminal and then back to one that shows Hyprland the internal laptop turns black.

logs:
[`Hyprland --sysinfo`](https://github.com/user-attachments/files/18151095/sysinfo.txt)

aquamarine 0.5.0:
[`hyprctlr monitors all -j`](https://github.com/user-attachments/files/18151040/monitors-working.json) 
[`hyprland.log`](https://github.com/user-attachments/files/18151045/hyprland-working.log)

aquamarine 0.5.1
[`hyprctlr monitors all -j` (initial state)](https://github.com/user-attachments/files/18151044/monitors-init.json)
[`hyprctlr monitors all -j` (after switching VTs)](https://github.com/user-attachments/files/18151042/monitors-switched.json)
[`hyprland.log`](https://github.com/user-attachments/files/18151046/hyprland-bug.log)

ikalco:
fixes hyprwm/Hyprland#8790

edit:
it think fixes #117 also

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -1429,11 +1429,6 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
         }
     }
 
-    if (STATE.drmFormat == DRM_FORMAT_INVALID) {
-        backend->backend->log(AQ_LOG_ERROR, "drm: No format for output");
-        return false;
-    }
-
     if (COMMITTED & COutputState::eOutputStateProperties::AQ_OUTPUT_STATE_FORMAT) {
         // verify the format is valid for the primary plane
         bool ok = false;
@@ -1448,6 +1443,11 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
             backend->backend->log(AQ_LOG_ERROR, "drm: Selected format is not supported by the primary KMS plane");
             return false;
         }
+    }
+
+    if (STATE.enabled && STATE.drmFormat == DRM_FORMAT_INVALID) {
+        backend->backend->log(AQ_LOG_ERROR, "drm: No format for output");
+        return false;
     }
 
     if (STATE.adaptiveSync && !connector->canDoVrr) {


vaxerski (commented):
how the fuck would format be invalid otherwise?

ikalco:
default is DRM_FORMAT_INVALID

so on first connect, rule says disabled, HL commits with no buffer or format set, which is fine since it's a disable commit
but it fails because the default is DRM_FORMAT_INVALID, so `drm: No format for output`

vaxerski:
shouldnt we check for a format then only if it's enabled and not only if format committed?

ikalco:
yeah your right

src/backend/drm/DRM.cpp
```diff
@@ -1427,11 +1427,10 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
             backend->backend->log(AQ_LOG_ERROR, "drm: No mode on enable commit");
             return false;
         }
-    }
-
-    if (STATE.drmFormat == DRM_FORMAT_INVALID) {
-        backend->backend->log(AQ_LOG_ERROR, "drm: No format for output");
-        return false;
+        if (STATE.drmFormat == DRM_FORMAT_INVALID) {
```
vaxerski, line 1430: now it wont be checked if we commit just format :P

I'd do it below the format commit

```cpp
if (STATE.enabled) {
   // check
}
```

ikalco: yeah im gonna uninstall linux now ;-;

vaxerski: lol

vaxerski (approved):
:+1:

PR #128, opened 2025-01-05
Revert #108

### LINKED ISSUE #125: Permanent black screen after waking up from sleep (s2idle)
The recent aquamarine update from v0.5.0 to v0.5.1 has resulted in a critical failure when my laptop wakes up from s2idle sleep mode. The screen is black and it seems Hyprland is unable to restart, since I cannot find any errors from systemd services activated before and after the suspend action.

After analyzing all packages installed on my system, I've identified the error originates from this aquamarine update. I've also bisected the commits in the update and localized the issue to 4468981c1c50999f315baa1508f0e53c4ee70c52 (see #108 for discussion about the change and #107 for the issue that prompted it).

From my understanding, this commit was intended to prevent a segfault on startup which was occurring on certain systems. I am not aware if this was ever an issue on my system, but I do know that #107 suggested the fix that was eventually applied in #108 as a potential workaround and noted that a more complete solution would involve identifying calls to `CDRMAtomicImpl::reset()` and ensuring that a null pointer will not be dereferenced after the call.

I am not sure how this change results in a sleep mode failure, but my guess is that upon waking up from the sleep state, the fix in #108 prevents critical logic from executing because of a potential null pointer dereference and Hyprland is unable to start.

quantum9Innovation:
This is causing a major regression on some Nvidia laptops that leads to the system becoming frozen after returning from suspend. Fixes #125.

This reverts commit 4468981c1c50999f315baa1508f0e53c4ee70c52 for the time being.

### DIFF
--- a/src/backend/drm/impl/Atomic.cpp
+++ b/src/backend/drm/impl/Atomic.cpp
@@ -162,9 +162,6 @@ bool Aquamarine::CDRMAtomicRequest::commit(uint32_t flagssss) {
         return false;
     }
 
-    if (!conn)
-        return false;
-
     if (auto ret = drmModeAtomicCommit(backend->gpu->fd, req, flagssss, &conn->pendingPageFlip); ret) {
         backend->log((flagssss & DRM_MODE_ATOMIC_TEST_ONLY) ? AQ_LOG_DEBUG : AQ_LOG_ERROR,
                      std::format("atomic drm request: failed to commit: {}, flags: {}", strerror(ret == -1 ? errno : -ret), flagsToStr(flagssss)));


vaxerski:
this is not valid, you are literally creating a segfault with this. There is no way this commit is a source of any regression.

youwen5:
Although reverting the commit was probably not the best fix, I can confirm that after bisecting the aquamarine tree, this exact commit is causing a sleep regression on at least a few systems. The bisection was performed on NixOS by precisely overriding the aquamarine source and building the rest of Hyprland, etc against it so I'm reasonably confident that there were no residual environmental reasons for the regression besides this exact commit.

quantum9Innovation:
Superceded by #149

PR #130, opened 2025-01-07
GPU hotplug support

maround95:
### Use Case

I have a Lenovo Legion 5 Pro (16ACH6H) laptop with an AMD iGPU and an NVIDIA dGPU. I want the flexibility to use different modules for the dGPU depending on my needs:

1. **No module**: startup and power-saving.
2. **`vfio-pci`**: For passthrough to a vm.
3. **`nouveau`**: To use the connectors connected to the dGPU exclusively for driving additional monitors.
4. **`nvidia` (without modesetting)**: For CUDA workloads, offloaded rendering, etc.

---

### Workflow involving Hyprland
1. **Load nouveau**:  
  I can now use the outputs connected to my dGPU only.
2. **Signal Aquamarine to release the dGPU**:
  ```bash
  udevadm trigger --type=devices --action=remove --subsystem-match=drm --property-match="MINOR=0"
  ```
3. **Unload nouveau and load something else.**

---

I'm happy to take this to the finish line in case any changes are required.

### DIFF
--- a/include/aquamarine/allocator/Allocator.hpp
+++ b/include/aquamarine/allocator/Allocator.hpp
@@ -26,6 +26,6 @@ namespace Aquamarine {
         virtual Hyprutils::Memory::CSharedPointer<CBackend> getBackend()                                                                                           = 0;
         virtual int                                         drmFD()                                                                                                = 0;
         virtual eAllocatorType                              type()                                                                                                 = 0;
-        virtual void                                        destroyBuffers() {};
+        virtual void                                        destroyBuffers();
     };
 };
--- a/include/aquamarine/allocator/Allocator.hpp
+++ b/include/aquamarine/allocator/Allocator.hpp
@@ -26,5 +26,6 @@ namespace Aquamarine {
         virtual Hyprutils::Memory::CSharedPointer<CBackend> getBackend()                                                                                           = 0;
         virtual int                                         drmFD()                                                                                                = 0;
         virtual eAllocatorType                              type()                                                                                                 = 0;
+        virtual void                                        destroyBuffers() {};
     };
 };
--- a/src/allocator/GBM.cpp
+++ b/src/allocator/GBM.cpp
@@ -228,14 +228,15 @@ Aquamarine::CGBMBuffer::CGBMBuffer(const SAllocatorBufferParams& params, Hypruti
 }
 
 Aquamarine::CGBMBuffer::~CGBMBuffer() {
+    for (size_t i = 0; i < (size_t)attrs.planes; i++)
+        close(attrs.fds.at(i));
+
     events.destroy.emit();
     if (bo) {
         if (gboMapping)
             gbm_bo_unmap(bo, gboMapping); // FIXME: is it needed before destroy?
         gbm_bo_destroy(bo);
     }
-    for (size_t i = 0; i < (size_t)attrs.planes; i++)
-        close(attrs.fds.at(i));
 }
 
 eBufferCapability Aquamarine::CGBMBuffer::caps() {
@@ -280,9 +281,23 @@ void Aquamarine::CGBMBuffer::endDataPtr() {
     }
 }
 
+void CGBMAllocator::destroyBuffers() {
+    for (auto &buf : buffers) {
+        buf.impl_->destroy();
+    }
+}
+
 CGBMAllocator::~CGBMAllocator() {
-    if (gbmDevice)
-        gbm_device_destroy(gbmDevice);
+    if (!gbmDevice)
+        return;
+
+    int fd = gbm_device_get_fd(gbmDevice);
+    gbm_device_destroy(gbmDevice);
+
+    if (fd < 0)
+        return;
+
+    close(fd);
 }
 
 SP<CGBMAllocator> Aquamarine::CGBMAllocator::create(int drmfd_, Hyprutils::Memory::CWeakPointer<CBackend> backend_) {
--- a/src/allocator/GBM.cpp
+++ b/src/allocator/GBM.cpp
@@ -284,7 +284,7 @@ void Aquamarine::CGBMBuffer::endDataPtr() {
 
 void CGBMAllocator::destroyBuffers() {
     for (auto& buf : buffers) {
-        buf.reset()
+        buf.reset();
     }
 }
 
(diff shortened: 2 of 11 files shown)


include/aquamarine/allocator/Allocator.hpp
```diff
@@ -26,5 +26,6 @@ namespace Aquamarine {
         virtual Hyprutils::Memory::CSharedPointer<CBackend> getBackend()                                                                                           = 0;
         virtual int                                         drmFD()                                                                                                = 0;
         virtual eAllocatorType                              type()                                                                                                 = 0;
+        virtual void                                        destroyBuffers() {};
```
vaxerski, line 29: please put the definition in the cpp file, here just declare it.

maround95: I get a compilation error if I don't define this in Allocator.hpp (IAllocator), as opposed to GBM.cpp where I could get rid of the definition.
I'm not a cpp programmer, so I might be missing something.

vaxerski: drop the {} and define it in Allocator.cpp, I can do it for you if you don't know how to

maround95:
> looking at some of the stuff changed like closing fds, this might fix/help #83

This patch also allows me to dissociate Hyprland from nouveau if I launch Hyprland with nouveau loaded from the start. I tested it with AQ_DRM_DEVICES=/dev/dri/card1(amd):/dev/dri/card0(nvidia)

vaxerski:
lgtm, can you please clang-format all the files though?

maround95:
Done

src/backend/drm/DRM.cpp
```diff
@@ -269,7 +306,16 @@ std::vector<SP<CDRMBackend>> Aquamarine::CDRMBackend::attempt(SP<CBackend> backe
 }
 
 Aquamarine::CDRMBackend::~CDRMBackend() {
-    ;
+    for (auto conn : connectors) {
+        if (conn && conn->output)
+            conn->output->events.destroy.emit();
+        conn.impl_->destroy();
```
vaxerski, line 312: wait, wtf is this? These pointers are managed, don't do this.

just conn.reset()

src/backend/drm/DRM.cpp
```diff
@@ -269,7 +306,16 @@ std::vector<SP<CDRMBackend>> Aquamarine::CDRMBackend::attempt(SP<CBackend> backe
...
+    }
+
+    rendererState.allocator->destroyBuffers();
+
+    rendererState.renderer.impl_->destroy();
+    rendererState.allocator.impl_->destroy();
```
vaxerski, line 318: same here

maround95:
I forgot about having this hack because connectors were not being destroyed, so I forced them to. What I wanted was conn.disconnect() to destroy the outputs and then the conn refcount works out.

src/allocator/GBM.cpp
```diff
@@ -228,14 +228,15 @@ Aquamarine::CGBMBuffer::CGBMBuffer(const SAllocatorBufferParams& params, Hypruti
 }
 
 Aquamarine::CGBMBuffer::~CGBMBuffer() {
+    for (size_t i = 0; i < (size_t)attrs.planes; i++)
```
vaxerski, line 231: {} please, only no {} for ifs

src/backend/Backend.cpp
```diff
@@ -271,6 +271,27 @@ void Aquamarine::CBackend::dispatchIdle() {
     updateIdleTimer();
 }
 
+void Aquamarine::CBackend::onNewGpu(std::string path) {
+    const auto primary    = std::ranges::find_if(implementations, [](SP<IBackendImplementation> value) { return value->type() == Aquamarine::AQ_BACKEND_DRM; });
+    const auto primaryDrm = primary != implementations.end() ? ((Aquamarine::CDRMBackend*)(*primary).get())->self.lock() : nullptr;
+
+    auto       ref = CDRMBackend::fromGpu(path, self.lock(), primaryDrm);
+    if (ref == nullptr) {
```
vaxerski, line 279: !ref

vaxerski (commented):
some final style nits. Content lgtm

vaxerski (approved):
thanks!

KZDKM:
> This patch also allows me to dissociate Hyprland from nouveau if I launch Hyprland with nouveau loaded from the start. I tested it with AQ_DRM_DEVICES=/dev/dri/card1(amd):/dev/dri/card0(nvidia) to set amd as the primary drm backend.

Do you have to set your iGPU as the primary drm backend to hotplug the dGPU? I am currently running an AMD iGPU+dGPU setup with card0(dGPU) set to primary via the envvar. Hyprland crashes when I release it to pass to vfio-pci while keeping the iGPU on host.

maround95:
> Do you have to set your iGPU as the primary drm backend to hotplug the dGPU? I am currently running an AMD iGPU+dGPU setup with card0(dGPU) set to primary via the envvar. Hyprland crashes when I release it to pass to vfio-pci while keeping the iGPU on host.

Your usecase is similar to mine. You can't release the primary drm without Hyprland crashing.

You should have the iGPU as your primary in this case. You can then release hyprland resources from the dGPU with the udevadm command above, rmmod nouveau and bind vfio-pci for passthrough.

I do not have to set AQ_DRM_DEVICES because I have configured the dGPU to be off before login so the iGPU is always the primary, and I can then load the module I want for the dGPU.
If you wish both to be on before login, I would set AQ_DRM_DEVICED, with the iGPU first (primary) followed by the dGPU (so that you can use the dGPU connectors if you wish).

KZDKM:
I wonder if forcing iGPU for drm would impact gaming performance on linux. I am on a desktop so ideally I should be running everything directly off dGPU whenever possible. Regardless, this is something I wanted for a long while so thank you for your PR!

navneetankur:
Does this also work for nvidia drivers, or only for nouveau. When hyprland starts with nvidia driver present. After running:
`udevadm trigger --type=devices --action=remove --subsystem-match=drm --property-match="MINOR=0"`
Or:
`udevadm trigger --type=devices --action=remove --subsystem-match=drm --attr-match="vendor=0x10de"`
Hyprland is still using /dev/nvidia0, and unbind does not work until closing hyprland.

Additionally: AQ_DRM_DEVICES, does not have nvidia card present.

PR #133, opened 2025-01-11
remove wrong property settings

UjinT34:
Missed some properties while moving to separate function. Those might incorrectly reset the state.


vaxerski (approved):
makes sense.

PR #135, opened 2025-01-21
Core: add clang-tidy

Honkazel:
`.clang-tidy` stolen from Hyprland, yeah
Is there any ABI break?

### DIFF
--- a/src/backend/drm/Renderer.cpp
+++ b/src/backend/drm/Renderer.cpp
@@ -213,7 +213,7 @@ SP<CDRMRenderer> CDRMRenderer::attempt(Hyprutils::Memory::CSharedPointer<CGBMAll
     renderer->backend         = backend_;
     gBackend                  = backend_;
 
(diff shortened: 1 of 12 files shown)


src/backend/drm/Renderer.cpp
```diff
@@ -99,7 +99,7 @@ void main() {
 
 inline void loadGLProc(void* pProc, const char* name) {
     void* proc = (void*)eglGetProcAddress(name);
-    if (proc == NULL) {
+    if (proc == nullptr) {
```
vaxerski, line 102: !proc if we're at it.

vaxerski (commented):
no ABI break, ABI can't break from .cpp files :P

src/backend/drm/Renderer.cpp
Honkazel: Done

vaxerski (approved):
thanks!

PR #138, opened 2025-01-24
consolidate EGL init/state

ikalco:
<s>next step is to use CEGLRenderer in HL</s>

reasoning: egl init, loading, and egl specific funcs (createEGLImage) are all in one place

<s>note:
egl queried formats don't have external modifiers stripped here, but hl impl does, so will deal with it in hl pr</s>

### DIFF
--- a/include/aquamarine/allocator/DRMDumb.hpp
+++ b/include/aquamarine/allocator/DRMDumb.hpp
@@ -63,6 +63,6 @@ namespace Aquamarine {
         int                                                          drmfd = -1;
 
         friend class CDRMDumbBuffer;
-        friend class CDRMRenderer;
+        friend class CEGLRenderer;
     };
 };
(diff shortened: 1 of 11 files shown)


ikalco:
completely forgot about this lol
my original thought was that egl init would all be in one place and it could possibly allow for using multiple renderers in hl later or maybe lead to implementing different renderers under a parent class

but i don't want to do all that right now, so abi can be broken then if me or someone else ever tries

after removing the abi breaking stuff, I would say this is just an overall improvement
this should also fix hyprwm/Hyprland#9239 (it crashes in Renderer dtor, prob from eglTerminate on null)

vaxerski (commented):
technically you break CRenderer abi but is that even used by hl directly?

ikalco:
uuh CDRMRenderer from aq? no HL doesn't use that directly
my original thought was to move some of the egl init in CHyprOpenGLImpl to aq

vaxerski:
no, dont

vaxerski (approved):

PR #140, opened 2025-02-01
Support setting drm content type property

UjinT34:
API to set drm content type property. https://drmdb.emersion.fr/properties/3233857728/content%20type
Nvidia doesn't support it. Untested. Probably needs monitor support to be useful.


vaxerski:
what would it even do

vaxerski (approved):

PR #141, opened 2025-02-02
Compiler fixes and options

Honkazel:
As i said in the title
There's only 5 warnings left
3 of them is unused variable, 1 unused, but set variable and a implicit fallthrough


vaxerski (approved):
thanks :)

PR #143, opened 2025-02-02
drm: only allow multigpu blit to implcit, not all external formats

ikalco:
realized while looking back at #114, oops


vaxerski (commented):
since when in god's name is it legal to use implicit modifiers across gpus

ikalco:
i belive its like this
oes external image/texture -> implicit format framebuffer

idk if its written down anywhere but it fixed the original issue in #114
and its part of what that pr did
> allow the blit to_buffer to use external formats (cause implicit is external)

vaxerski (approved):
ok

PR #146, opened 2025-02-23
Get max bpc from drm format

UjinT34:
Should fix `bitdepth,10` for mesa drivers. https://github.com/hyprwm/Hyprland/issues/9103
Untested.
Assumes that when `data.mainFB` is present it has a dma buffer attached.


vaxerski (approved):

Candyhands:
I was one of the people who had an issue with 10 bit not working, this PR fixed it for me.

Issue #134, opened 2025-01-18
hardware cursors on dgpu fails to blit


gulafaran:
not sure if this should be reported here or at hyprland, but turning on hw cursors with or without use_cpu_buffers makes it fail blitting and endlessly spam the log such as. i do have an external monitor attached to the nvidia gpu.

```
 [LOG] [AQ] DRM Dumb: Allocated a new buffer with primeFD 99, size [Vector2D: x: 256, y: 256] and format AR24                                
 [LOG] [AQ] DRM Dumb: Allocated a new buffer with primeFD 208, size [Vector2D: x: 256, y: 256] and format AR24                               
 [LOG] [AQ] Swapchain: Reconfigured a swapchain to [Vector2D: x: 256, y: 256] AR24 of length 2                                               
 [ERR] [AQ] GBM: Allocating with modifiers and flags failed, falling back to modifiers without flags                                         
 [LOG] [AQ] GBM: Allocated a new buffer with size [Vector2D: x: 256, y: 256] and format AR24 with modifier 0 aka LINEAR                      
 [ERR] [AQ] [GLES] Error in call at 592@Renderer.cpp: 0x502                                                                                  
 [ERR] [AQ] [GLES] Error in call at 609@Renderer.cpp: 0x506                                                                                  
 [ERR] [AQ] GBM: Allocating with modifiers and flags failed, falling back to modifiers without flags                                         
 [LOG] [AQ] GBM: Allocated a new buffer with size [Vector2D: x: 256, y: 256] and format AR24 with modifier 0 aka LINEAR                      
 [ERR] [AQ] [GLES] Error in call at 592@Renderer.cpp: 0x502                                                                                  
 [ERR] [AQ] [GLES] Error in call at 609@Renderer.cpp: 0x506                                                                                  
 [LOG] [AQ] Swapchain: Reconfigured a swapchain to [Vector2D: x: 256, y: 256] AR24 of length 2                                               
 [LOG] [AQ] EGL (blit): No attachment in from, creating a new image                                                                          
 [LOG] [AQ] CDRMRenderer::glTex: found format+mod, external = true                                                                           
 [LOG] [AQ] EGL (blit): No attachment in to, creating a new image                                                                            
 [ERR] [AQ] [GLES] Error in call at 693@Renderer.cpp: 0x502                                                                                  
 [ERR] [AQ] EGL (blit): glCheckFramebufferStatus failed: 0                                                                                   
[...]

PR #147, opened 2025-02-24
Fix multi-GPU support for Nvidia

PlasmaPower:
This PR has two commits to fix running aquamarine against multiple Nvidia GPUs which provide different output monitors:
- A small commit to not force linear allocation on devices which don't support it, even if multigpu is enabled. This is necessary as Nvidia doesn't support the linear format.
- A Renderer commit to, if creating an EGL image from the source buffer fails, instead copy it to an intermediate buffer on the CPU memory first, then copy that intermediate buffer to the output buffer. This fixes Nvidia multigpu because Nvidia doesn't support attaching another gpu's buffer as an EGL image, so instead this PR first reads the source buffer into CPU memory using the primary renderer's EGL context, and then does the rest of the blit as normal on the secondary GPU which writes to the output buffer.

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -1700,8 +1700,12 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
             }
 
             auto NEWAQBUF   = mgpu.swapchain->next(nullptr);
+            SP<Aquamarine::CDRMRenderer> primaryRenderer;
+            if (backend->primary) {
+                primaryRenderer = backend->primary->rendererState.renderer;
+            }
             auto blitResult = backend->rendererState.renderer->blit(
-                STATE.buffer, NEWAQBUF, (COMMITTED & COutputState::eOutputStateProperties::AQ_OUTPUT_STATE_EXPLICIT_IN_FENCE) ? STATE.explicitInFence : -1);
+                STATE.buffer, NEWAQBUF, primaryRenderer, (COMMITTED & COutputState::eOutputStateProperties::AQ_OUTPUT_STATE_EXPLICIT_IN_FENCE) ? STATE.explicitInFence : -1);
             if (!blitResult.success) {
                 backend->backend->log(AQ_LOG_ERROR, "drm: Backend requires blit, but blit failed");
                 return false;
@@ -1882,7 +1886,11 @@ bool Aquamarine::CDRMOutput::setCursor(SP<IBuffer> buffer, const Vector2D& hotsp
             }
 
             auto NEWAQBUF = mgpu.cursorSwapchain->next(nullptr);
-            if (!backend->rendererState.renderer->blit(buffer, NEWAQBUF).success) {
+            SP<Aquamarine::CDRMRenderer> primaryRenderer;
+            if (backend->primary) {
+                primaryRenderer = backend->primary->rendererState.renderer;
+            }
+            if (!backend->rendererState.renderer->blit(buffer, NEWAQBUF, primaryRenderer).success) {
                 backend->backend->log(AQ_LOG_ERROR, "drm: Backend requires blit, but cursor blit failed");
                 return false;
             }
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -1699,11 +1699,10 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
                 return false;
             }
 
-            auto NEWAQBUF   = mgpu.swapchain->next(nullptr);
+            auto                         NEWAQBUF = mgpu.swapchain->next(nullptr);
             SP<Aquamarine::CDRMRenderer> primaryRenderer;
-            if (backend->primary) {
+            if (backend->primary)
                 primaryRenderer = backend->primary->rendererState.renderer;
-            }
             auto blitResult = backend->rendererState.renderer->blit(
                 STATE.buffer, NEWAQBUF, primaryRenderer, (COMMITTED & COutputState::eOutputStateProperties::AQ_OUTPUT_STATE_EXPLICIT_IN_FENCE) ? STATE.explicitInFence : -1);
             if (!blitResult.success) {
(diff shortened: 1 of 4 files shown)


vaxerski:
I am a bit torn on the second one. That copy is very slow. Will it not lag into oblivion?

src/backend/drm/Renderer.hpp
```diff
@@ -24,18 +24,22 @@ namespace Aquamarine {
...
-        GLuint                                        fbo = 0, rbo = 0;
-        SGLTex                                        tex;
-        Hyprutils::Signal::CHyprSignalListener        bufferDestroy;
+        EGLImageKHR                            eglImage = nullptr;
+        GLuint                                 fbo = 0, rbo = 0;
+        SGLTex                                 tex;
+        Hyprutils::Signal::CHyprSignalListener bufferDestroy;
+        // This is malloc'd manually instead of using a std::vector to keep lifetime management in line with the rest of the class,
+        // which e.g. doesn't immediately free the eglImage on drop but instead waits until onBufferAttachmentDrop.
+        uint8_t*                                      intermediateBuf    = nullptr;
```
vaxerski, line 41: that's completely unnecessary. You can use a vector and reset it in the drop() handler.

src/backend/drm/Renderer.cpp
```diff
@@ -698,6 +698,52 @@ SGLTex CDRMRenderer::glTex(Hyprutils::Memory::CSharedPointer<IBuffer> buffa) {
     return tex;
 }
 
+// TODO: Would another format be more efficient than GL_RGBA?
+const GLenum PIXEL_BUFFER_FORMAT = GL_RGBA;
```
vaxerski, line 702: constexpr

vaxerski:
> It manages 55 fps on a 4k display on my system which is not great, not terrible. Mutter does the same copy from what I can see and I'm not aware of an alternative.

alright.

src/backend/drm/Renderer.hpp
PlasmaPower: I was worried about it getting freed early, not freed late. But yeah I can just replace this with a vector. I don't think I need to reset it in the drop handler because that's only called from the destructor of the buffer which should call the destructor of the attachments immediately afterwards.

PlasmaPower: I've gone ahead and refactored this to use std::vector and std::span for the intermediateBuf. I'm coming from rust so let me know if any of this isn't idiomatic and I'd be happy to fix it (e.g. I debated using a std::optional for intermediateBuf but I figured it'd be better to just check if it's empty in C++).

src/backend/drm/DRM.cpp
```diff
@@ -1700,8 +1700,12 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
             }
 
             auto NEWAQBUF   = mgpu.swapchain->next(nullptr);
+            SP<Aquamarine::CDRMRenderer> primaryRenderer;
+            if (backend->primary) {
```
vaxerski, line 1704: no {}

src/backend/drm/DRM.cpp
```diff
@@ -1882,7 +1886,11 @@ bool Aquamarine::CDRMOutput::setCursor(SP<IBuffer> buffer, const Vector2D& hotsp
             }
 
             auto NEWAQBUF = mgpu.cursorSwapchain->next(nullptr);
-            if (!backend->rendererState.renderer->blit(buffer, NEWAQBUF).success) {
+            SP<Aquamarine::CDRMRenderer> primaryRenderer;
+            if (backend->primary) {
```
vaxerski, line 1890: same

src/backend/drm/Renderer.cpp
```diff
@@ -9,6 +9,7 @@
 #include "Shared.hpp"
 #include "FormatUtils.hpp"
 #include <aquamarine/allocator/GBM.hpp>
+#include <span>
```
vaxerski, line 12: you should include this in the header and not the source file

src/backend/drm/Renderer.cpp
```diff
@@ -846,22 +893,40 @@ CDRMRenderer::SBlitResult CDRMRenderer::blit(SP<IBuffer> from, SP<IBuffer> to, i
...
+
             // should never remove anything, but JIC. We'll leak an EGLImage if this removes anything.
             from->attachments.removeByType(AQ_ATTACHMENT_DRM_RENDERER_DATA);
-            from->attachments.add(makeShared<CDRMRendererBufferAttachment>(self, from, nullptr, 0, 0, fromTex));
+            from->attachments.add(newAttachment);
+        }
+
+        if (!intermediateBuf.empty()) {
+            // Note: this might modify from's attachments
+            primaryRenderer->readBuffer(from, intermediateBuf);
```
vaxerski, line 929: needs a null check. primaryRenderer is not guaranteed, and I remember some odd configurations where it is null

src/backend/drm/Renderer.cpp
```diff
@@ -698,6 +699,52 @@ SGLTex CDRMRenderer::glTex(Hyprutils::Memory::CSharedPointer<IBuffer> buffa) {
     return tex;
 }
 
+// TODO: Would another format be more efficient than GL_RGBA?
```
vaxerski, line 702: no, all suck.

src/backend/drm/Renderer.cpp
PlasmaPower: primaryRenderer can definitely be null, but intermediateBuf should never be populated without a primary renderer. That said, absolutely better safe than sorry. I've added the null check

ikalco:
this should also fix #134 by using the copy path in your second point

PlasmaPower:
I don't have the hardware to test it but I don't think it'll fix that one actually. In my case (two Nvidia GPUs) it was failing to create the `from` buffer, but in the linked issue (an iGPU as primary + an Nvidia dGPU as secondary it sounds like?) it looks like it was failing to call `glEGLImageTargetRenderbufferStorageOES` on the `to` buffer from the logs. I'm not sure which Renderer blit is being called on in that case and which GPUs the buffers belong to, so I'm not sure what the correct fix for that issue would be, but it would probably build on the infrastructure in this PR.

PlasmaPower:
Fixed the merge conflict with #150

vaxerski (approved):
thanks!

PR #149, opened 2025-02-27
fix atomic reset being skipped, check nullptr elsewhere instead

### LINKED ISSUE #125: Permanent black screen after waking up from sleep (s2idle)

ikalco:
kinda reverts 4468981c, really just fixes it better

probably fixes #125
also fixes issue I observed where hw pointer isn't reset on vt switch so you can get a cursor on multiple screens

edit:
btw the `&conn->pendingPageFlip` paramter is just a classic `void* data` and its ok to set null

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -909,7 +909,7 @@ int Aquamarine::CDRMBackend::drmFD() {
 static void handlePF(int fd, unsigned seq, unsigned tv_sec, unsigned tv_usec, unsigned crtc_id, void* data) {
     auto pageFlip = (SDRMPageFlip*)data;
 
-    if (!pageFlip->connector)
+    if (!pageFlip || !pageFlip->connector)
         return;
 
     pageFlip->connector->isPageFlipPending = false;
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
thanks!

PR #150, opened 2025-02-27
drm: Use RAII to manage EGL context

PlasmaPower:
Using RAII to manage the EGL context:
- Correctly restores the EGL context if an early return occurs due to an error
- Correctly handles recursive Renderer calls (though we don't do any at the moment, #147 only calls a separate Renderer)
- Prevents other code oversights like forgetting to call restoreEGL()

I took inspiration from the std::lock_guard API for this, but I haven't worked in C++ for a while so let me know if there's a better API.

This has a small conflict with #147 but I can resolve that depending on which one gets merged first.

### DIFF
--- a/src/backend/drm/Renderer.cpp
+++ b/src/backend/drm/Renderer.cpp
@@ -447,7 +447,7 @@ void CDRMRenderer::initContext(bool GLES2) {
             backend->log(AQ_LOG_DEBUG, "CDRMRenderer: Got a high priority context");
     }
 
-    EglContextGuard eglContext(*this);
+    CEglContextGuard eglContext(*this);
 
     EGLEXTENSIONS = (const char*)glGetString(GL_EXTENSIONS);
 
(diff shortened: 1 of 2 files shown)


src/backend/drm/Renderer.cpp
```diff
@@ -575,25 +571,25 @@ SP<CDRMRenderer> CDRMRenderer::attempt(SP<CBackend> backend_, Hyprutils::Memory:
     return renderer;
 }
 
-void CDRMRenderer::setEGL() {
+EglContextGuard::EglContextGuard(CDRMRenderer& renderer_) : renderer(renderer_) {
```
vaxerski, line 574: can be const ref

vaxerski (commented):
generally lgtm

vaxerski (commented):
thanks

vaxerski (approved):
sorry for the delay, thanks!

PlasmaPower:
No worries at all, thank you for all your work reviewing PRs and maintaining Hyprland!

PR #151, opened 2025-03-02
backend: Fix Wayland backend idleCallbacks destruction

PlasmaPower:
The old code was crashing for me on shutdown due to the fact that `CWaylandBackend`'s reverse field order destruction destroys `idleCallbacks` before destroying `outputs`, meaning that when `~CWaylandOutput` calls `backend->idleCallbacks.clear();` it's clearing an already destroyed vector which crashes.

To fix this I just removed the FIXME hack and properly handled the potential `sendFrameAndSetCallback` UAF by using the weak pointer to check if the output was already destroyed or not.

### DIFF
--- a/src/backend/Wayland.cpp
+++ b/src/backend/Wayland.cpp
@@ -508,7 +508,6 @@ Aquamarine::CWaylandOutput::CWaylandOutput(const std::string& name_, Hyprutils::
 }
 
 Aquamarine::CWaylandOutput::~CWaylandOutput() {
-    backend->idleCallbacks.clear(); // FIXME: mega hack to avoid a UAF in frame events
     events.destroy.emit();
     if (waylandState.xdgToplevel)
         waylandState.xdgToplevel->sendDestroy();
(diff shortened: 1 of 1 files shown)


vaxerski:
why not just `if (backend)` then?

PlasmaPower:
You mean checking that when `idleCallbacks.clear()` is called? I thought I might as well fix the FIXME while I was at it, though yeah checking `!backend.expired()` works (the expired check is necessary because this'll be called by the destructor of the backend so `backend` is still truthy). If you mean checking `if (backend)` in the callback, the concern there is that the output might be destroyed before the backend is I believe, and then the idle callback would be referencing an invalid output when called.

vaxerski (approved):
ah, I re-read it and noticed that github folded the diff weirdly. Now that makes sense, yes.

src/backend/Wayland.cpp
```diff
@@ -769,8 +768,13 @@ void Aquamarine::CWaylandOutput::scheduleFrame(const scheduleFrameReason reason)
 
     if (waylandState.frameCallback)
         frameScheduledWhileWaiting = true;
-    else
-        backend->idleCallbacks.emplace_back([this]() { sendFrameAndSetCallback(); });
+    else {
+        auto w = self;
+        backend->idleCallbacks.emplace_back([w]() {
```
vaxerski, line 773: actually you dont need that line above, you can do `[w = self]` it should do the same, though I am not 100% sure if it will use a & or not. Worth checking

PlasmaPower: It copies as it should. I wasn't aware of that syntax when I wrote this but it works well

vaxerski (approved):
thanks!

PR #152, opened 2025-03-02
misc: Add strict typing to attachment manager

PlasmaPower:
Feel free to reject this PR if the new abstraction is too complicated or otherwise not your preference, but it fixes a couple of potential pitfalls and simplifies the external API:
- You no longer have to manage the eAttachmentType and figure out which enum values correspond to which types. This also makes it easier to add new attachment types in the future, and provides flexibility to use this in other non-DRM contexts or with external attachment types.
- You no longer have to cast returned attachments to the right type, as the attachment manager now does the cast for you. This also avoids the potential pitfall of casting to the wrong attachment type.
- Previously, we called `removeByType` before `add` because it was possible to have multiple attachments of the same type by accident, but attachments are now stored in a map instead of a vector to guarantee uniqueness.

### DIFF
--- a/include/aquamarine/misc/Attachment.hpp
+++ b/include/aquamarine/misc/Attachment.hpp
@@ -21,13 +21,10 @@ namespace Aquamarine {
     // However, only one attachment of a given type is permitted.
     class CAttachmentManager {
       public:
-        // has checks if the manager has an attachment of the specified type
         template <AttachmentConcept T>
         bool has() const {
             return attachments.contains(typeid(T));
         }
-        // get retrieves the attachment on the specified type,
-        // returning nullptr if no attachment of the specified type is present.
         template <AttachmentConcept T>
         const Hyprutils::Memory::CSharedPointer<T>& get() const {
             auto it = attachments.find(typeid(T));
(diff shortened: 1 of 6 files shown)


include/aquamarine/misc/Attachment.hpp
```diff
@@ -2,33 +2,56 @@
...
+        void add(Hyprutils::Memory::CSharedPointer<IAttachment> attachment);
+        // remove removes the specified attachment, doing nothing if it does not exist.
+        // Note that this will not remove a different attachment of the same type.
+        void remove(Hyprutils::Memory::CSharedPointer<IAttachment> attachment);
+        // remove removes the attachment of the specified type, doing nothing if it does not exist
+        template <AttachmentConcept T>
+        void removeByType() {
+            attachments.erase(typeid(T));
+        }
+        // clear removes all attachments
```
vaxerski, line 51:
```cpp
// eats a banana
void eatBanana();
```

most of the comments here are useless clutter, if not all. Function names here are self-documenting.

vaxerski (commented):
generally fine, depends on RTTI but we already do in many places

include/aquamarine/misc/Attachment.hpp
PlasmaPower: Makes sense, I've removed them but kept a comment on add about how it removes any previous attachment of the same type because that might be an unknown edge case to the consumer otherwise.

vaxerski (approved):
thanks!

PR #157, opened 2025-03-11
buffer: Get rid of buffer locking

PlasmaPower:
Also changes the public fields in IBuffer to getters, so that they can be overridden.


PlasmaPower:
Based on conversations in discord about double-commits and reviewing the overall architecture here, I no longer think RAII is the way to go for buffer locking and we should make do with the backendRelease system instead

PR #159, opened 2025-03-16
backend, allocator: Fix multi-GPU format selection

PlasmaPower:
Swapchains were created with the primary allocator, but the output's backend. This meant that the wrong format would be selected with multi-GPU setups such as NVIDIA primary and Intel secondary. We still try to use the LINEAR format if available, but if it's not available, we should use the format from the primary GPU instead of the secondary GPU.

My previous PR https://github.com/hyprwm/aquamarine/pull/147 fixed having two NVIDIA GPUs or having Intel as primary and NVIDIA as secondary, but this PR fixes having NVIDIA as primary and an Intel as secondary. Previously, the swapchain would create an Intel format buffer to render to for the secondary monitor, which NVIDIA would then fail to read due to being an unsupported format. Now, it'll correctly create an NVIDIA format buffer for the secondary monitor, which blit will then copy over to the Intel GPU memory, transforming the formats in the process.

This PR also adds a getPrimary method to IBackendImplementation for use in Hyprland to make the same fix for the cursor swapchain there: https://github.com/hyprwm/Hyprland/pull/9645

### DIFF
--- a/include/aquamarine/backend/Backend.hpp
+++ b/include/aquamarine/backend/Backend.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <hyprutils/memory/SharedPtr.hpp>
+#include <hyprutils/memory/WeakPtr.hpp>
 #include <hyprutils/signal/Signal.hpp>
 #include <vector>
 #include <functional>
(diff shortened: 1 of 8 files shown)


PlasmaPower:
This also is confirmed to fix an NVIDIA primary + Intel iGPU secondary setup for at least one user on Discord :)

vaxerski (approved):
thanks!

PR #161, opened 2025-03-26
drm: Add Vulkan host mapping blit

PlasmaPower:
This PR adds a more efficient alternative CPU based copy if the formats between the buffers are the same (i.e. there's two NVIDIA cards). In addition, this PR lays the groundwork for other uses of Vulkan especially in blit. Vulkan is a lot more powerful than OpenGL here, and any future blit optimizations will likely also use it.

There are two reasons this copy is more efficient than the `intermediateBuf` codepath:
1. There's no format conversion. OpenGL needs to convert the pixel data into `GL_RGBA` whereas Vulkan can keep the pixel data in the existing format and simply copy the memory. However, the Vulkan code could be extended to support a format conversion in the future.
2. The copy is asynchronous. The previous OpenGL copy was entirely blocking, relying on glReadPixels and glTexImage2d. This new copy uses Vulkan semaphores and fences to consume and produce a sync_file, and even has a background thread for the memcpy itself to make it asynchronous (don't worry, I made sure it's isolated and doesn't cause concurrency concerns for the rest of the system).

Right now I think this is only useful if you have two NVIDIA cards, but @gulafaran has been looking at using Vulkan to optimize the blit for the NVIDIA<>Intel case and this PR lays important groundwork for that as well.

If you want to test this PR on a single-GPU system you can use this patch to force enable Vulkan based blitting: https://gist.github.com/PlasmaPower/1499e2db2abbb1ab88aa1334c145f3fa

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,8 @@ pkg_check_modules(
   gbm
   libudev
   libdisplay-info
-  hwdata)
+  hwdata
+  vulkan)
 
 configure_file(aquamarine.pc.in aquamarine.pc @ONLY)
 
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -16,6 +16,7 @@
   pkg-config,
   seatd,
   udev,
+  vulkan-loader,
   wayland,
   wayland-protocols,
   wayland-scanner,
@@ -62,6 +63,7 @@ in
       pixman
       seatd
       udev
+      vulkan-loader
       wayland
       wayland-protocols
       wayland-scanner
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -1679,7 +1679,7 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
 
         SP<CDRMFB> drmFB;
 
-        if (backend->shouldBlit() || true) {
+        if (backend->shouldBlit()) {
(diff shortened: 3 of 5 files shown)


PlasmaPower:
I was just going to comment on the Vulkan dep, thanks @fufexan! I think we can make it a build dep only and use entirely dynamic library loading at runtime to fail gracefully on a lack of Vulkan, but at this point I think pretty much everyone has Vulkan installed on their system.

vaxerski:
oh god vulkan

@outfoxxed has multi amd thats worth testing too

outfoxxed:
> oh god vulkan
> 
> @outfoxxed has multi amd thats worth testing too

Lmk what I need to test and I can do that.

PlasmaPower:
Could you first test if the latest aquamarine release (v0.8.0) is still working with displays connected to each of your AMD GPUs, and then test if this PR breaks it if it is working?

outfoxxed:
Both 0.8.0 and this branch work as expected. Anything beyond functioning at surface level that needs to be tested?

PlasmaPower:
Nope, just that the displays show up and aren't super slow or anything.

LuckShiba:
Any updates on this? I have an Intel + NVIDIA setup and the constant CPU high usage on the external display (that I think that is caused by the CPU copy) is very annoying, this may help but I can't really test right now

PR #163, opened 2025-03-29
Restore hdr metadata after vt switch

UjinT34:
Should fix https://github.com/hyprwm/Hyprland/issues/9724
Couldn't replicate the issue. HDR settings are checked before every commit. The only way this can happen is when `output->state->state().hdrMetadata` doesn't match the real drm state.


vaxerski (approved):
lgtm

PR #165, opened 2025-04-17
DRM: Drop KMS FBs on disconnect and disable.

vaxerski:
Drop the refs when we disable the crtc or disconnect the connector

ref https://github.com/hyprwm/Hyprland/issues/10098

### DIFF
--- a/include/aquamarine/backend/DRM.hpp
+++ b/include/aquamarine/backend/DRM.hpp
@@ -107,6 +107,7 @@ namespace Aquamarine {
 
     struct SDRMPlane {
         bool                                         init(drmModePlane* plane);
+        void                                         dropFBs();
 
         uint64_t                                     type      = 0;
         uint32_t                                     id        = 0;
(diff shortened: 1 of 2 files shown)


vaxerski:
@rkben

PlasmaPower:
Not that this is a bad change, but are the `SDRMPlane`s also being leaked? Why isn't the destructor cleaning these up?

vaxerski:
IIRC planes can't change at runtime

PlasmaPower:
If we assign to them, then shouldn't we only be leaking one set of buffers per plane? Or is there just a large enough number of planes that the memory leak appears anyways?

rkben:
Issue persists with https://github.com/hyprwm/Hyprland/commit/225e13c3cc83308175f0a9aa18cfa31324155034 built against this.

vaxerski:
I guess I need to look harder tomorrow (3am rn)

PR #172, opened 2025-05-19
a few small renderer optimisations

gulafaran:
dont call glflush when its not needed, avoid vector reallocations in waitOnSync and guard glgeterror behind AQ_TRACE


vaxerski (approved):
thanks!

PR #173, opened 2025-05-20
depend on gles3 and use vertex array objects

gulafaran:
begin depend on gles3 and above, and use vertex array objects when blitting, ensure same program doesnt get called with gluseprogram more then once.

### DIFF
--- a/src/backend/drm/Renderer.cpp
+++ b/src/backend/drm/Renderer.cpp
@@ -249,6 +249,13 @@ std::optional<std::vector<std::pair<uint64_t, bool>>> CDRMRenderer::getModsForFo
     return result;
 }
 
+void CDRMRenderer::useProgram(GLuint prog) {
+    if (m_currentProgram == prog)
+        return;
+
+    GLCALL(glUseProgram(prog));
+}
+
 bool CDRMRenderer::initDRMFormats() {
     std::vector<EGLint> formats;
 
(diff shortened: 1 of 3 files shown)


src/backend/drm/Renderer.hpp
```diff
@@ -86,8 +88,57 @@ namespace Aquamarine {
 
         struct {
             struct SShader {
+                ~SShader() {
```
vaxerski, line 91: these aren't templated. Move them to the source.

src/backend/drm/Renderer.cpp
```diff
@@ -402,27 +457,15 @@ void CDRMRenderer::initContext(bool GLES2) {
...
-        attrs.push_back(2);
-        attrs.push_back(EGL_CONTEXT_MINOR_VERSION);
-        attrs.push_back(0);
-    } else {
-        attrs.push_back(EGL_CONTEXT_MAJOR_VERSION);
-        attrs.push_back(3);
-        attrs.push_back(EGL_CONTEXT_MINOR_VERSION);
-        attrs.push_back(2);
-    }
+    attrs.push_back(EGL_CONTEXT_MAJOR_VERSION);
```
vaxerski, line 460: Use 3.0. 3.0 systems exist and arent that uncommon. You don't seem to use anything from 3.1/3.2

gulafaran: the if else case down below drops it to 3.0 if context fails, as it is on -git too, but sure i can change it to only 3.0 if you want

vaxerski (approved):

PR #175, opened 2025-05-23
reduce a bit of local temporar copies going around in renderer

gulafaran:
use const refs where applicable, make sgltex a uniqueptr and weakptr reference it.


vaxerski (approved):
thanks!

PR #183, opened 2025-07-03
Prevent DRM Backend from modesetting unnecessarily

### LINKED ISSUE #181: Screen turns off and then turns on when starting aquamarine
When aquamarine is started, instead of the contents of whatever the app is drawing just instantly appearing, the screen turns off and only turns on 1 second later. I guess this is probably due to aquamarine sending a null commit or a commit without a buffer when it detects a screen the first time.

My proposed solution to this is to have an option when initializing aquamarine which tells it to never send null commits and only send commits with `glClear(0.0, 0.0, 0.0, 0.0)`'ed buffer which I think would prevent the screen from turning off.

For hyprland this might be fine but for my case I need it to have a contentful paint as fast as possible

coffeeispower:
Closes #181

### DIFF
--- a/src/backend/drm/impl/Atomic.cpp
+++ b/src/backend/drm/impl/Atomic.cpp
@@ -100,8 +100,16 @@ void Aquamarine::CDRMAtomicRequest::addConnector(Hyprutils::Memory::CSharedPoint
     TRACE(backend->log(AQ_LOG_TRACE, std::format("atomic addConnector values: CRTC {}, mode {}", enable ? connector->crtc->id : 0, data.atomic.modeBlob)));
 
     conn = connector;
+    drmModeModeInfo* currentMode = connector->getCurrentMode();
+    bool modeDiffers = true;
+    if (currentMode) {
+        modeDiffers = memcmp(currentMode, &data.modeInfo, sizeof(drmModeModeInfo)) != 0;
+        free(currentMode);
+    }
 
-    addConnectorModeset(connector, data);
+    if (modeDiffers) {
+        addConnectorModeset(connector, data);
+    }
     addConnectorCursor(connector, data);
 
     add(connector->id, connector->props.crtc_id, enable ? connector->crtc->id : 0);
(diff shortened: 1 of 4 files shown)


src/backend/drm/impl/Atomic.cpp
```diff
@@ -100,8 +100,16 @@ void Aquamarine::CDRMAtomicRequest::addConnector(Hyprutils::Memory::CSharedPoint
...
     conn = connector;
+    drmModeModeInfo* currentMode = connector->getCurrentMode();
+    bool modeDiffers = true;
+    if (currentMode) {
+        modeDiffers = memcmp(currentMode, &data.modeInfo, sizeof(drmModeModeInfo)) != 0;
+        free(currentMode);
+    }
 
-    addConnectorModeset(connector, data);
+    if (modeDiffers) {
```
vaxerski, line 110: no {} around short ifs

src/backend/drm/impl/Atomic.cpp
```diff
@@ -455,22 +463,7 @@ bool Aquamarine::CDRMAtomicImpl::commit(Hyprutils::Memory::CSharedPointer<SDRMCo
...
-    for (auto const& conn : backend->connectors) {
-        request.add(conn->id, conn->props.crtc_id, 0);
-    }
-
-    for (auto const& plane : backend->planes) {
-        request.planeProps(plane, nullptr, 0, {});
-    }
-
-    return request.commit(DRM_MODE_ATOMIC_ALLOW_MODESET);
+    return true;
```
vaxerski, line 466: are we sure about this? Where is reset called? Maybe we should reset sometimes?

src/backend/drm/impl/Atomic.cpp
coffeeispower: oh sorry, i'm used to rust where you can't omit the braces, will fix

src/backend/drm/impl/Atomic.cpp
vaxerski: ok

vaxerski (commented):
c-f needed

vaxerski (approved):

PR #193, opened 2025-07-18
core: introduce drmRenderNodeFD()

gulafaran:
make a getter for drmRenderNodeFD() so we can create sync timelines on the rendernode on devices not supporting fences on the card* device directly.

add the rendernode matching the drm card if there is one. if none exist fallback to first found.

### DIFF
--- a/src/backend/Session.cpp
+++ b/src/backend/Session.cpp
@@ -165,13 +165,8 @@ void Aquamarine::CSessionDevice::resolveMatchingRenderNode(struct udev_device* c
     if (!cardDevice)
         return;
 
-    auto pciParent = udev_device_get_parent_with_subsystem_devtype(cardDevice, "pci", nullptr);
-    if (!pciParent)
-        return;
-
-    const char* pciSyspath = udev_device_get_syspath(pciParent);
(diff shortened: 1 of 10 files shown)


src/backend/Session.cpp
```diff
@@ -157,6 +161,84 @@ bool Aquamarine::CSessionDevice::supportsKMS() {
...
+
+    struct udev_enumerate* enumerate = udev_enumerate_new(session->udevHandle);
+    if (!enumerate)
+        return;
+
+    udev_enumerate_add_match_subsystem(enumerate, "drm");
+    udev_enumerate_scan_devices(enumerate);
+
+    struct udev_list_entry* devices = udev_enumerate_get_list_entry(enumerate);
+    struct udev_list_entry* entry   = nullptr;
```
vaxerski, line 179: why are all those `struct` keywords here? this is not C?

vaxerski (commented):
generally lgtm

src/backend/Session.cpp
gulafaran: changed.

vaxerski (approved):

PR #197, opened 2025-07-21
allocator: Default to forcing linear blit for multigpu

### LINKED ISSUE #196: Multi-GPU rendering causes stutters due to b058847
When Hyprland draws on both monitors simultaneously, stutters can be observed. I've bisected it to https://github.com/hyprwm/aquamarine/commit/b058847592e5fb3f121ef909601ab635247a0d1c.

I have an AMD+AMD gpu configuration, each monitor on a separate GPU. The stuttering has been observed on the primary gpu which Hyprland is configured to run with. I cannot test the reverse as the other gpu cannot manage vsync.

Reverting that commit on top of aq-git and applying the following patch to make hyprland-git compile completely fixes the stuttering:
```diff
diff --git a/src/managers/PointerManager.cpp b/src/managers/PointerManager.cpp
index 7d6f58e0..4212661d 100644
--- a/src/managers/PointerManager.cpp
+++ b/src/managers/PointerManager.cpp
@@ -423,8 +423,8 @@ SP<Aquamarine::IBuffer> CPointerManager::renderHWCursorBuffer(SP<CPointerManager
             }
 
             auto backend                      = state->monitor->m_output->getBackend();
-            auto primary                      = backend->getPrimary();
-            state->monitor->m_cursorSwapchain = Aquamarine::CSwapchain::create(allocator, primary ? primary.lock() : backend);
+            //auto primary                      = backend->getPrimary();
+            state->monitor->m_cursorSwapchain = Aquamarine::CSwapchain::create(allocator, backend);// primary ? primary.lock() : backend);
         }
 
         auto options     = state->monitor->m_cursorSwapchain->currentOptions();
```

This has been tested with osu-lazer on the primary gpu+monitor, and the stutters can be observed in the game's debug HUD if direct scanout is on (moreso than with it off at least). The secondary monitor on the secondary gpu is running glxgears.

Before reverting the commit:

Image

After reverting the commit:

Image

PlasmaPower:
Makes the behavior introduced in #174 default. I've confirmed that the fallback in case linear allocation fails means that this still works on NVIDIA<>NVIDIA. I left the env around in case it needs disabled, but since the new behavior seems to work on both NVIDIA<>NVIDIA and AMD<>AMD I think we should be good.

Fixes #196 by forcing linear allocation to avoid CPU blitting on AMD<>AMD.

### DIFF
--- a/src/allocator/GBM.cpp
+++ b/src/allocator/GBM.cpp
@@ -140,7 +140,7 @@ Aquamarine::CGBMBuffer::CGBMBuffer(const SAllocatorBufferParams& params, Hypruti
         return;
     }
 
-    static const auto forceLinearBlit = envEnabled("AQ_FORCE_LINEAR_BLIT");
+    static const auto forceLinearBlit = !envExplicitlyDisabled("AQ_FORCE_LINEAR_BLIT");
     auto const        oldMods         = explicitModifiers; // used in FORCE_LINEAR_BLIT case.
     if (MULTIGPU && !forceLinearBlit) {
         // Try to use the linear format if available for cross-GPU compatibility.
--- a/src/include/Shared.hpp
+++ b/src/include/Shared.hpp
@@ -6,6 +6,7 @@
 
 namespace Aquamarine {
     bool envEnabled(const std::string& env);
+    bool envExplicitlyDisabled(const std::string& env);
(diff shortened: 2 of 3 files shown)


vaxerski (approved):
rawr, needs a wiki mr

PR #184, opened 2025-07-04
Avoid unnecessary modeset (fixed)

coffeeispower:
This PR adds back fastboot to aquamarine but `hyprctl dpms` is now fixed.

PR #198, opened 2025-07-25
Fix hdr modeset

UjinT34:
#184 broke hdr by introducing an extra condition. #188 fixed that by moving hdr stuff outside of `addConnectorModeset`. This skips `data.modeset` check for hdr properties. Probably causes black screen on some systems and increased cpu usage on others.

Restored the old hdr modeset behavior while keeping faster modeset from #184. HDR should work as expected, faster modeset and black screen issue need more testing.

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -8,6 +8,7 @@ Unless specified otherwise, a variable is enabled if and only if it's set to `1`
 `AQ_NO_ATOMIC` -> Disables drm atomic modesetting
 `AQ_MGPU_NO_EXPLICIT` -> Disables explicit syncing on mgpu buffers
 `AQ_NO_MODIFIERS` -> Disables modifiers for DRM buffers
+`AQ_FAST_MODESET` -> Skips modeset if mode is unchanged
 
 ### Debugging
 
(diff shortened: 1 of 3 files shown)


UjinT34:
https://discord.com/channels/961691461554950145/1070436481912549497/1398284251186860032

vaxerski:
yeah but can they test this MR is what I meant

vaxerski (commented):
this will cause modesets each frame, that's horrible...

m-bartlett:
> who had the black screen issues? Can we get them here?

hi I'm someone with the black screen issue caused by the commit in the screenshot, I independently arrived at the same commit https://github.com/hyprwm/aquamarine/commit/8b22613953961cbea21a620d0559c7ea27dc7e15 doing my own bisect in the thread here:

https://github.com/hyprwm/Hyprland/discussions/10248#discussioncomment-13825221

I just tried this PR's latest commit and I get a completely black screen without AQ_NO_ATOMIC=1 which is the same behavior as the latest commit and all releases after v0.9.1. However, simply doing a `git revert 8b226` on the latest commit https://github.com/hyprwm/aquamarine/commit/50637ed23e962f0db294d6b0ef534f37b144644b was enough to get the display working as expected without AQ_NO_ATOMIC.

PR #201, opened 2025-08-06
Update README.md to include installation command.

blucardin:
A small PR to include the installation command in the README. Improves ease-of-use, especially for beginners trying to install hyprland.


vaxerski:
Absolutely agreed

PR #202, opened 2025-08-10
renderer: use rendernode if available

gulafaran:
init the CDRMRenderer on the rendernode if available otherwise fallback to the displaynode.

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -602,7 +602,7 @@ bool Aquamarine::CDRMBackend::initMgpu() {
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


gulafaran:
question is if even the allocators should perhaps be on rendernodes aswell, in general only place where the display node should be used is when modesetting/kms comes into the picture

gulafaran:
> question is if even the allocators should perhaps be on rendernodes aswell, in general only place where the display node should be used is when modesetting/kms comes into the picture

seems `int Aquamarine::CBackend::reopenDRMNode(int drmFD, bool allowRenderNode)` should already be doing so

vaxerski (approved):
halal brother

PR #203, opened 2025-08-17
drmdumb: reopen displaynode

gulafaran:
if the backends now gets a rendernode it wont call the authmagic on the displaynode making dumb buffer creation failing on certain drivers. so reopen it again in the creation of the CDRMDumbAllocator.

im not sure if this breaks it further, or actually solves it. i cant even reproduce it. the only place so far i noticed this happening is on the nix test suite on hyprland. https://github.com/hyprwm/Hyprland/actions/runs/17022533814/job/48253900013?pr=11087#step:5:5414

ive tested intel only, nvidia only, and intel + nvidia. so idk


vaxerski (approved):
lgtm lets ball

PR #204, opened 2025-08-19
render: revert #203 and half of #202

gulafaran:
either there is a fd going wrong here or certain drivers just plain fails with this. revert for now. but leave the cdrmrenderer using rendernode. 

cant reproduce it locally but CI running llvmpipe and virtualmachine fails on hyprland.


vaxerski:
does it fix the CI though?

Issue #210, opened 2025-09-22
Black screen on startup with DisplayLink HDMI monitor when using Aquamarine backend on Hyprland


keros-Guardia:
### **Description:**

I am experiencing a black screen issue when connecting a DisplayLink HDMI monitor on Hyprland at startup. The problem only occurs if the monitor is plugged in before launching Hyprland. If I connect the DisplayLink monitor after Hyprland has started, it works correctly without any issues.

### **Steps to reproduce:**

1. Connect a DisplayLink HDMI monitor to the system.

2. Start Hyprland with the DisplayLink monitor already connected.

3. Observe that the external monitor remains black.

4. Unplug and replug the DisplayLink monitor after Hyprland has started.

5. The monitor then works as expected.

### **Expected behavior:**

The DisplayLink HDMI monitor should be detected and initialized correctly by Hyprland and the Aquamarine rendering backend on startup, without needing to unplug and replug it.

### **Actual behavior:**

When the DisplayLink monitor is connected at startup, the screen remains black until the monitor is unplugged and reconnected after Hyprland has launched.

### **Additional details:**

The problem seems to be related to the Aquamarine rendering backend library (aquamarine package version 0.9.4-2 on Arch Linux).

Downgrading aquamarine to version 0.9.3-1 fails due to unresolved dependency issues (libdisplay-info.so=2-64).

The DisplayLink drivers and services are installed, enabled, and up to date.

The issue is reproducible consistently.

Workaround is to connect the DisplayLink monitor after Hyprland startup.

Running aquamarine as a standalone command is not applicable as it is a rendering library, not an executable.

### **System information:**

Distro: Arch Linux (or derivative)

Hyprland version: [insert your version here]

Aquamarine version: 0.9.4-2

DisplayLink adapter: USB to HDMI

Kernel version: 6.16.8-zen2-1-zen

GPU: NVIDIA Corporation GA106M [GeForce RTX 3060 Mobile / Max-Q] (rev a1)

Dock : Bus 002 Device 004: ID 17e9:6006 DisplayLink Dell Universal Dock D6000

### **Possible related issues:**

DisplayLink devices might not be correctly initialized by Aquamarine at startup.

Timing issues between DisplayLink driver initialization and Aquamarine backend startup.

### **Suggested solutions or workarounds:**

Delay Aquamarine or Hyprland startup until DisplayLink services are fully ready.

Provide better runtime detection and reinitialization of DisplayLink devices in Aquamarine.

[...]

PR #214, opened 2025-10-04
rendernode: dont bother finding one on evdi

gulafaran:
evdi/displaylink devices doesnt give us a rendernode at all, and asahi was hard to associate if a rendernode was related to the displaynode/card so we fallback to just first found if that occurs. might be more appropiate to figure out a proper way to deal with these cases then workaround ever single driver that doesnt give rendernodes but so far seems only to be evdi.

problem that occured is a 3 gpu situation, the evdi card ended up using a rendernode for a complete different gpu because it fallbacked to that.

fixes: #210


vaxerski (approved):
thanks!

PR #221, opened 2025-11-28
backend: add freeDrmDevice on error in initDmabuf

zacoons:
I was just perusing the code for nefarious reasons and found this, which may cause a small memory leak in some rare cases.


vaxerski (approved):
thanks

PR #223, opened 2025-12-13
Remove Unnecessary Modeset After Restoring from VT

Nosamdaman:
This PR removes a forced monitor mode reset after restoring from a VT switch.

I've tested DPMS and different monitor modes with this change, and everything seems to be working exactly as expected.

This PR has a sister in Hyprland [here](https://github.com/hyprwm/Hyprland/pull/12645)

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -383,9 +383,6 @@ void Aquamarine::CDRMBackend::restoreAfterVT() {
 
     backend->log(AQ_LOG_DEBUG, "drm: Rescanned connectors");
 
-    if (!impl->reset())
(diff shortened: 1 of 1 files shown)


Nosamdaman:
> are we sure this is unnecessary? I've had many MRs to "remove unnecessary modesets" and 90% of them broke something.

We are not, but I've read through many of the previous MR's and they mention broken DPMS, which seems to be working for me on both an NVIDIA and Intel system.

Furthermore, this change in particular only disables the explicit reset when restoring from a VT. If that were to cause issues, why is it that no modeset is performed when Hyprland initially takes over from a text mode TTY? If you have any suggestions about how I might test this further, please let me know.

PR #224, opened 2025-12-20
output: ensure only changed things commit

gulafaran:
avoid committing if nothing really changed.

#TODO
- [x] ensure fast modesetting still works
- [x] restore state if external drm master has changed driver state ("sort of done, requires hyprland PR aswell")
- [ ] does buffers need to be restored in restorevt?

### DIFF
--- a/include/aquamarine/output/Output.hpp
+++ b/include/aquamarine/output/Output.hpp
@@ -76,10 +76,10 @@ namespace Aquamarine {
             uint32_t                                       drmFormat = DRM_FORMAT_INVALID;
             Hyprutils::Memory::CSharedPointer<IBuffer>     buffer;
             int32_t                                        explicitInFence = -1, explicitOutFence = -1;
-            Hyprutils::Math::Mat3x3                        ctm;
+            Hyprutils::Math::Mat3x3                        ctm            = Hyprutils::Math::Mat3x3::identity();
             bool                                           hasCtm         = false;
             bool                                           wideColorGamut = false;
-            hdr_output_metadata                            hdrMetadata;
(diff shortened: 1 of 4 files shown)


gulafaran:
ok unless something shows up as broken when testing this is ready as is.

ive tried going through the entire flow and internalState is now matching driver state as long as we are the drm master, test commits or even failing commits doesnt call `COutputState::onCommit()` so the internalState becomes the "pending" state up until next successful commit where its applied and matching again. it now only commits things where the state has actually changed. and upon monitor connect we recreate current state so later commits can reset it, if needed.

the only slight change here is vtswitching might modeset more now then before and i see no clear way out of it. because the crtc props mismatch the driverstate on restoreAfterVT() so cant safely compare them without somehow reloading the connector/drm props somehow.

but the gain is that hdr and states is properly reset if an external actor has changed it and we either start HL or vtswitch to it.

Nosamdaman:
I've just updated to Hyprland 0.53.0 and confirmed that I still see this issue on this branch.

Nosamdaman:
> what does `hyprctl monitors | grep currentFormat` say before and after it becomes oversaturated

It says: `currentFormat: XRGB8888` both before and after the oversaturation.

GodTamIt:
N.B. for context, I had a related issue posted here: https://github.com/hyprwm/Hyprland/discussions/13036

Happy to test changes. Just ping me.

fxzzi:
been running for like over a month straight now, most stuff looks good for me. i can launch hyprland from tty on a 10 bit monitor without requiring a modeset.

there were a few things i noted and that tom is aware of but thought i'd share here:

- when using HDR, going to tty and back to hyprland and everything becomes fried. can be fixed just by toggling off and on HDR
- the first time you go to a TTY causes a modeset. any further switches between hyprland and tty don't modeset.

PR #227, opened 2025-12-29
libinput: fix pad groups not being ref'd

### LINKED ISSUE #207: Crash when unplugging Wacom tablet using Hyprland
Here is the log:

```
...
[LOG] [AQ] [libinput] event24 - Wacom Intuos Pro M Pad: device removed
```

[hyprland.log](https://github.com/user-attachments/files/22243835/hyprland.log)
[hyprlandCrashReport7764.txt](https://github.com/user-attachments/files/22296110/hyprlandCrashReport7764.txt)

The only relevant configuration I have is this:
```
device {
    name = wacom-intuos-pro-m-pen
    output = DP-2
}
```

The crash also happens without it.

Hyprland discussion: https://github.com/hyprwm/Hyprland/discussions/11641

DarkCreepy59:
Without refing pad groups they would get destroyed when the current event gets out of scope, causing a segfault when destroying a CLibinputTabletPad.

I can't test myself but it seems it fixes the issue.
Fixes #207


vaxerski (approved):
thanks!

PR #229, opened 2025-12-30
atomic fixes

gulafaran:
took parts out of the #224 so it can be both easier reviewed properly figured out

add missing hdr blob to apply/rollback, clamp the bpc to min/max advertised hw support and not only the format. and properly set the blobid to zero instead of setting the local copied ptr nullptr.

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -1436,6 +1436,9 @@ void Aquamarine::SDRMConnector::connect(drmModeConnector* connector) {
     if (!currentModeInfo && fallbackMode)
         output->state->setMode(fallbackMode);
 
+    if (currentModeInfo)
+        free(currentModeInfo);
+
     output->physicalSize = {(double)connector->mmWidth, (double)connector->mmHeight};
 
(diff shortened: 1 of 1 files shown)


src/backend/drm/DRM.cpp
```diff
@@ -1459,7 +1459,7 @@ void Aquamarine::SDRMConnector::connect(drmModeConnector* connector) {
 
     maxBpcBounds.fill(0);
 
-    if (props.values.max_bpc && !introspectDRMPropRange(backend->gpu->fd, props.values.max_bpc, maxBpcBounds.data(), &maxBpcBounds[1]))
+    if (props.values.max_bpc && !introspectDRMPropRange(backend->gpu->fd, props.values.max_bpc, &maxBpcBounds[0], &maxBpcBounds[1]))
```
vaxerski: this change does nothing and is uglier

gulafaran: i blame my history of C but yeah okay, changed back.

gulafaran:
> It's worth noting that the issue I saw with over-saturation when switching to and from a text tty in #224 doesn't occur on this branch.

yeah its the state changes, i just havent figured out what its messing up. so im moving over the bits i know is safe to lessen the changes in the other PR

PR #235, opened 2026-02-01
drm: use parent device matching for render nodes

who53:
Instead of blindly picking first node when the driver is non-pci or doesnt have any render node, this will resolve render node FD correctly.
Also remove EVDI check since it wont be needed after this change.


gulafaran:
> > does this work on those pesky arm devices?
> 
> no it wont if you want to associate render node with card3
> 
> I can think of only one solution we detect the subsystem of node if its platform/pci, we can set renderNodeFd to -1 if no renderNode (i have similar driver to evdi and it gets wrong render node since its non pci) for rest we fallback to first render node like before this should resolve correctly, no?

Yea might work

PR #241, opened 2026-02-11
drm: make pageflips drive the frames

gulafaran:
make the pageflips drive the frames instead of waiting on the idle dispatch, it adds unnecessery latency at the beginning of each frame.


vaxerski:
I dont understand what this changes

vaxerski (approved):
k

Issue #59, opened 2024-08-20
modesetting failing on external monitor


maaarghk:
Since hyprland 0.43 / aquamarine I can't use my external monitor anymore. its an LG Ultrafine which is a 5k tiling monitor. I don't know my jargon particularly well but I believe it exposes 2xPHYs, to do 5k mode you set both to 2560x2880, OR, you can just disable DP-1 and set DP-2 to 3840x2160 and it'll upscale it for you. which is what I do because I got fucked off with having to manually build wlroots with the tiling patch applied all the time.

I think maybe there are several issues, firstly, since aqua, DP-1 is not getting disabled according to i915_display_info (even though it is not showing up in hyprctl monitors), when I've done this before the monitor just decides it is misconfigured and displays nothing. 

Also between 0.3.3 and git, currentFormat goes from `XRGB8888` to `A XBGR8888 H Invalid`.  on wlroots it uses XRGB8888 and works fine. 

But either way I can get it into several broken states and can't seem to make an image come up. Here's a log with tracing enabled, and various i915_display_info dumps - wlroots one works, term one mirrors e-DP1, the rest are various kinds of not working on aquamarine.

[hyprland.log](https://github.com/user-attachments/files/16680680/hyprland.log)

[i915_display_info_hypr_wlroots.txt](https://github.com/user-attachments/files/16680688/i915_display_info_hypr_wlroots.txt)
[i915_display_info_on_term.txt](https://github.com/user-attachments/files/16680691/i915_display_info_on_term.txt)
[i915_display_info_hypr_aqua4.txt](https://github.com/user-attachments/files/16680698/i915_display_info_hypr_aqua4.txt)
[i915_display_info_on_hypr_aqua3.txt](https://github.com/user-attachments/files/16680699/i915_display_info_on_hypr_aqua3.txt)
[i915_display_info_on_hypr_aqua2.txt](https://github.com/user-attachments/files/16680700/i915_display_info_on_hypr_aqua2.txt)
[i915_display_info_on_hypr_aqua.txt](https://github.com/user-attachments/files/16680702/i915_display_info_on_hypr_aqua.txt)

maaarghk:
oh, and, an attempt to make search / google eventually work for future travellers -> `[ERR] [AQ] atomic drm request: failed to commit: Invalid argument, flags: ATOMIC_ALLOW_MODESET PAGE_FLIP_EVENT `

vaxerski:
0.43 doesn't exist xd

[...]

Issue #40, opened 2024-08-05
External Monitor stays just lit black (HDMI) Notebook Intel integrated + Nvidia but should work on Intel


romanstingler:
From the AQ PR my outstanding MultiMonitor Bug. 

I am using `Tag: v0.41.2-48-gf642fb97, commits: 4934` for now, as this is my office laptop, there everything is fine,
`DP-3` is my external HDMI monitor which stays black

I clean-built `hyprland-git  0.41.2.r166.0e86808e-1` the issue is still present. 
```
hyprcursor-git 0.1.9.r9.g4493a972-1
hypridle-git 0.1.2.r10.ge5366d34-1
hyprland-git  0.41.2.r166.0e86808e-1
hyprlang-git 0.5.2.r8.gadbefbf4-1
hyprlock-git 0.4.1.r3.gcf0e975f-1
hyprutils-git 0.2.1.r1.g0252fd13-1
hyprwayland-scanner-git 0.4.0.r4.ga048a6cb-1
```

```
~ ❯ printenv                                                                                       21:25
XDG_VTNR=1
SHLVL=2
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
SHELL=/usr/bin/fish
MEMORY_PRESSURE_WRITE=c29tZSAyMDAwMDAgMjAwMDAwMAA=
EDITOR=vim
LOGNAME=roman
ELECTRON_USE_OZONE=1
GDK_BACKEND=wayland
XDG_SESSION_ID=1
MOZ_ENABLE_WAYLAND=1
PWD=/home/roman
OZONE_PLATFORM=wayland
DISPLAY=:0
HYPRCURSOR_SIZE=24
XDG_BACKEND=wayland
MOTD_SHOWN=pam
MEMORY_PRESSURE_WATCH=/sys/fs/cgroup/system.slice/system-getty.slice/getty@tty1.service/memory.pressure
SUDO_RS_IS_UNSTABLE=I accept that my system may break unexpectedly
LANG=en_US.UTF-8
CREDENTIALS_DIRECTORY=/run/credentials/getty@tty1.service
XDG_CURRENT_DESKTOP=Hyprland
AQ_MGPU_NO_EXPLICIT=1
ELECTRON_OZONE_PLATFORM_HINT=auto
USER=roman
[...]

PR #254, opened 2026-03-04
drm: CRTC starvation recovery + clear stale page-flip state after suspend

UncleJ4ck:
## Setup

I run a laptop (Intel HD 630 iGPU, Quadro M620 dGPU unused) with 3 external monitors (HDMI-A-1, DP-3, DP-5) and the built-in lid display (eDP-1). The HD 630 has exactly 3 CRTCs but 4 connectors. Since I force `AQ_DRM_DEVICES=/dev/dri/intel-igpu` to avoid the dGPU, all 4 connectors compete for those 3 CRTCs.

A `lid-manager.sh` daemon disables eDP-1 when the lid is closed or when all 3 externals are connected (`hyprctl keyword monitor "eDP-1,disable"`).

---

### The bug

When I boot with lid closed and all 3 externals connected, everything works. But if I unplug one external and replug it (or open the lid while only 2 externals are connected), the formerly-starved connector never gets a CRTC. The output stays dark. Only fix: restart Hyprland entirely.

What happens: all 3 CRTCs get claimed at boot. When a connector disconnects, its CRTC should be freed, but the cleanup in `recheckCRTCs()` only *logged* "removing old crtc" without calling `crtc.reset()`, so the stale reference stayed. When the compositor disables an output, nothing triggers CRTC reassignment. `applyCommit()` updates `enabledState` but never tells the backend to re-evaluate which connectors should get which CRTCs.

This affects anyone with more connectors than CRTCs. Common on Intel iGPU laptops with docking stations. Related issues: aquamarine#36, #40, #59, Hyprland#7572, #7694.

### The fix

Three changes in `DRM.cpp`:

1. Two-pass CRTC assignment in `recheckCRTCs()`. Pass 1 releases CRTCs from disabled-but-connected outputs and assigns CRTCs to enabled connectors. Pass 2 gives remaining free CRTCs to disabled connectors as backup slots for quick re-enable. The disconnected-connector cleanup now actually calls `crtc.reset()`.

2. No-CRTC guard in `recheckOutputs()`. If a connector has no CRTC after `recheckCRTCs()`, `connect()` is skipped. It gets connected later when a CRTC frees up via the deferred recheck.

3. Deferred recheck on `enabledState` change in `applyCommit()`. When `enabledState` transitions, `addIdleEvent()` schedules `recheckOutputs()` on the next event loop iteration. Same mechanism already used for frame scheduling. Avoids reentrancy since `applyCommit` runs inside the DRM commit path.

### Safety

If all outputs have CRTCs (the common case), Pass 1 releases nothing, Pass 2 has no leftovers, and the idle-event never fires. Tiled display connectors (`tilingRedundant`) are skipped at the top of `recheckCRTCs()`. Multi-GPU setups run each backend independently. The idle event uses a weak pointer so it's safe if the backend is destroyed between scheduling and firing.

---

### The bug

After S3 suspend or suspend-then-hibernate, all monitors go permanently black. The only recovery is a hard restart. I ran into this on the same laptop after suspending overnight.

The page-flip lifecycle works like this: compositor submits a buffer, atomic commit sets `DRM_MODE_PAGE_FLIP_EVENT`, `isPageFlipPending` becomes true. Kernel completes the flip, delivers an event on the DRM fd, `handlePF()` fires and sets `isPageFlipPending = false`. Next frame gets scheduled.

During suspend, the display hardware powers off. Any in-flight page-flip completion event is lost because no vblank interrupt fires when the hardware is off. On resume, libseat fires `changeActive`, `restoreAfterVT()` runs, but `isPageFlipPending` is still true from before suspend. `handlePF()` never fires because there's no event to process. `scheduleFrame()` checks `isPageFlipPending`, sees true, returns early. No frames ever get scheduled. Display stays black.

VT switch doesn't hit this because the kernel preserves DRM state and queues pending page-flip events in the fd buffer. They get delivered when the session reactivates. During suspend the hardware is off, so there's nothing to deliver.

The regression trace: commit d83c97f (Dec 2025) removed `impl->reset()` from `restoreAfterVT()`. That reset used to deactivate all CRTCs at the kernel level, implicitly clearing stale page-flip state. It was removed for valid reasons (caused flicker during VT switch), but left no replacement for the suspend path. Commit 603f5cd (Feb 2026) added `isFrameRunning` as a third guard in `scheduleFrame()`, making the stale-state problem worse.

wlroots had the same bug class: issues #2290, #2325, #2373, #2395, fixed in commit 324eeaa0cd ("disable all CRTCs after VT switch").

Related: Hyprland#8312 ("freezes after suspension"), Hyprland#6289 ("black screen after resuming from suspend").

### The fix

Changes across `DRM.cpp`, `DRM.hpp`, `Atomic.cpp`, `Legacy.cpp`:

1. In `restoreAfterVT()`, clear `isPageFlipPending`, `isFrameRunning`, and `frameEventScheduled` for all connectors before `recheckOutputs()`. For VT switch this is a no-op since events still arrive and `handlePF()` would set them false anyway. For suspend it removes the blocker so frames can be scheduled again.

2. In `commitState()`, detect stale page-flips using a `CLOCK_BOOTTIME` timestamp recorded when `isPageFlipPending` is set. If a modeset commit finds a flip pending for >500ms (well past any vblank interval even at low refresh rates), treat it as stale and clear the flags. `CLOCK_BOOTTIME` is used instead of `CLOCK_MONOTONIC` because `CLOCK_MONOTONIC` does not advance during suspend on Linux, making elapsed time appear near-zero after resume.

3. Added `pageFlipPendingAtMs` field to `SDRMConnector`. Timestamp recorded in both atomic (`Atomic.cpp`) and legacy (`Legacy.cpp`) commit paths when `isPageFlipPending` is set true.

### Safety

**VT switch**: Pending page-flip events survive in the DRM fd buffer. After `restoreAfterVT()` returns, `dispatchEvents()` processes them. `handlePF()` fires and sees `isPageFlipPending = false` (already cleared). The `= false` is a no-op. `onPresent()` rotates buffer refs normally. No visible difference.

**S3/S4 resume**: No pending events in fd buffer. The flag clearing removes the blocker. The blocking modeset in `restoreAfterVT()` displays the restored frame. Next input or DPMS-on triggers `scheduleFrame()`, which succeeds, and frames render.

**Edge case — flip completes just before suspend**: `handlePF()` already fired before suspend. `isPageFlipPending` is already false. The clearing loop skips that connector. No change.

### Tested

On Intel HD 630 (3 CRTCs, 4 connectors) with:
- Short S3 suspend/resume: screens recover, log shows "Clearing stale page-flip state for HDMI-A-1 during modeset (pending for 503ms)"
- Long suspend-then-hibernate (>10h): screens recover on resume
- VT switch (Ctrl+Alt+F2 and back): no regression, screens restore cleanly
- Hotplug cycles after resume: CRTC starvation recovery still works
- Fresh boot: no false stale detection, flags are zero

### DIFF
--- a/include/aquamarine/backend/DRM.hpp
+++ b/include/aquamarine/backend/DRM.hpp
@@ -312,7 +312,8 @@ namespace Aquamarine {
         Hyprutils::Math::Vector2D                      cursorPos, cursorSize, cursorHotspot;
         Hyprutils::Memory::CSharedPointer<CDRMFB>      pendingCursorFB;
 
-        bool                                           isPageFlipPending = false;
+        bool                                           isPageFlipPending   = false;
+        uint64_t                                       pageFlipPendingAtMs = 0; // CLOCK_BOOTTIME ms when isPageFlipPending was set
         SDRMPageFlip                                   pendingPageFlip;
         bool                                           frameEventScheduled = false;
         bool                                           isFrameRunning      = false;
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -710,6 +710,12 @@ void Aquamarine::CDRMBackend::recheckCRTCs() {
             continue;
         }
 
+        // disabled outputs release their CRTCs so active outputs get priority
+        if (c->crtc && c->status == DRM_MODE_CONNECTED && c->output && !c->output->enabledState) {
+            backend->log(AQ_LOG_DEBUG, std::format("drm: {} is disabled, releasing crtc {}", c->szName, c->crtc->id));
+            c->crtc.reset();
+        }
+
         if (c->crtc && c->status == DRM_MODE_CONNECTED) {
             backend->log(AQ_LOG_DEBUG, std::format("drm: Skipping connector {}, has crtc {} and is connected", c->szName, c->crtc->id));
             continue;
@@ -739,12 +745,16 @@ void Aquamarine::CDRMBackend::recheckCRTCs() {
 
         bool assigned = false;
 
-        // try to use a connected connector
+        // try to use a connected, enabled connector
         for (auto const& c : recheck) {
             if (!(c->possibleCrtcs & (1 << i)))
                 continue;
 
             if (c->status != DRM_MODE_CONNECTED)
+                continue;
+
+            // Pass 1 only assigns to enabled connectors
+            if (c->output && !c->output->enabledState)
                 continue;
 
             // deactivate old output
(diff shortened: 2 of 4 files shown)


UncleJ4ck:
hey @vaxerski @gulafaran, any news on this? rebased on latest main, builds clean. also @ettancos is using it as a base in hyprwm/Hyprland#13828 for gpu reset recovery.

gulafaran:
> @gulafaran can you test this? I don't hw

"This affects anyone with more connectors than CRTCs." i dont have that. so besides looking at the code that seems reasonable, no clue.

my laptop only has igpu with internal panel, and dp/hdmi port goes to the nvidia card.

vaxerski (approved):

PR #273, opened 2026-04-10
backend: handle unhandled libinput switch enum with default case

B2krobbery:
The switch over libinput_event_switch_get_switch() does not handle all enum values,
which results in a compiler warning (-Wswitch).
Added a default case to ensure all enum values are handled without altering behavior.


vaxerski (approved):
tnx

Issue #272, opened 2026-04-09
Framework Desktop AMD (Strix Halo): Aquamarine/DRM crashes on connector teardown and render path under Hyprland


johngig1:
## System
- Hardware: Framework Desktop AMD Ryzen AI Max 300 Series
- GPU: AMD Strix Halo / Radeon 8050S/8060S (`1002:1586`)
- OS: Arch Linux / Omarchy
- Hyprland: `0.54.3`
- Aquamarine: `0.10.0`
- Mesa: `26.0.3`
- Kernel tested:
  - `linux`
  - `linux-lts`

## Symptoms
Frequent Hyprland crashes and earlier full freezes, often while using the mouse / scrolling in Chrome on X.com.

Originally the machine sometimes hard-froze and required reset. After switching to `linux-lts` and adding:
- `amdgpu.gpu_recovery=1`
- `amdgpu.vm_update_mode=3`
- `amdgpu.gttsize=8192`

the kernel-level symptoms improved, but Hyprland/Aquamarine crashes still occur.

## Crash signature 1: Aquamarine connector teardown SIGSEGV
This happened even in safe mode:
Hyprland --watchdog-fd 4 --safe-mode
coredumpctl / gdb shows:

Aquamarine::SDRMConnector::disconnect()
Aquamarine::SDRMConnector::~SDRMConnector()
Aquamarine::CDRMBackend::~CDRMBackend()

and the actual segfault is:
Aquamarine::CLogger::log(this=0x0, ...)

with message:
drm: Not disconnecting connector DP-1 because it's already disconnected

Relevant backtrace excerpt:

#0  Aquamarine::CLogger::log (this=0x0, level=Aquamarine::AQ_LOG_DEBUG, str="drm: Not disconnecting connector DP-1 because it's already disconnected")
#1  Aquamarine::SDRMConnector::disconnect
#2  Aquamarine::SDRMConnector::~SDRMConnector
#10 Aquamarine::CDRMBackend::~CDRMBackend
#24 Aquamarine::CBackend::~CBackend

### Crash signature 2: render-path SIGABRT

A newer crash also happens on HDMI, so this is not only DP-specific.

Backtrace:
#0  abort
#5  CHyprOpenGLImpl::begin
#6  CHyprRenderer::beginRender
#7  CHyprRenderer::renderMonitor
#8  CMonitorFrameScheduler::onFrame
#11 libaquamarine.so.9 + 0xad509
#12 drmHandleEvent
[...]

PR #274, opened 2026-04-11
drm: guard logging in connector disconnect to avoid null dereference

B2krobbery:
Related to #272

Adds a defensive check before logging in SDRMConnector::disconnect()
to avoid a potential null pointer dereference.

In some cases (e.g. after GPU reset or hotplug events), disconnect()
may be called when the connector is already disconnected. This can
lead to invalid backend/logger state.
This may occur during backend teardown, where connector disconnect
is triggered while backend state is being destroyed.
This change ensures backend pointers are valid before calling log(),
preventing a possible crash without altering behavior.
Tested by building aquamarine with Ninja and verifying correct usage
of CWeakPointer via .lock(). The change does not alter control flow
or cleanup logic, only guards logging in an already disconnected state.

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -1700,12 +1700,9 @@ void Aquamarine::SDRMConnector::connect(drmModeConnector* connector) {
 
 void Aquamarine::SDRMConnector::disconnect() {
     if (!output) {
-        if (backend) {
-            auto b = backend->backend.lock();
-            if (b)
-                b->log(AQ_LOG_DEBUG,
-                    std::format("drm: Not disconnecting connector {} because it's already disconnected", szName));
-        }
+        if (backend && backend->backend)
+            backend->backend->log(AQ_LOG_DEBUG,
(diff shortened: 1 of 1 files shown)


src/backend/drm/DRM.cpp
```diff
@@ -1700,7 +1700,12 @@ void Aquamarine::SDRMConnector::connect(drmModeConnector* connector) {
 
 void Aquamarine::SDRMConnector::disconnect() {
     if (!output) {
-        backend->backend->log(AQ_LOG_DEBUG, std::format("drm: Not disconnecting connector {} because it's already disconnected", szName));
+        if (backend) {
+            auto b = backend->backend.lock();
```
vaxerski: the lock here is redundant, also you should probably check for backend && backend->backend

B2krobbery: Got it, thanks I’ve updated the check to use backend && backend ->backend
and removed the lock().

vaxerski (approved):
lgtv

PR #275, opened 2026-04-11
fix(drm/atomic): retry DPMS ON commit on wake to unstick deep-sleep sinks

dottorblaster:
Some sinks, like NVIDIA Pascal at high refresh rates and AMD parts with GFXOFF, drop the first atomic commit that brings a connector out of DPMS off, leaving the monitor black until the user manually cycles DPMS again. This adds a bounded retry (100/250/500 ms backoff, 3 attempts) around CDRMAtomicImpl::commit that only kicks in on an actual wake transition.

If this is not ok and goes against the engineering principles of the project feel free to close this. :)

### DIFF
--- a/src/backend/drm/impl/Atomic.cpp
+++ b/src/backend/drm/impl/Atomic.cpp
@@ -1,7 +1,9 @@
 #include <aquamarine/backend/drm/Atomic.hpp>
+#include <chrono>
 #include <cstdint>
 #include <cstring>
 #include <drm_mode.h>
+#include <thread>
 #include <xf86drm.h>
 #include <xf86drmMode.h>
 #include <sys/mman.h>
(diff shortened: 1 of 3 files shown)


src/backend/drm/impl/Atomic.cpp
```diff
@@ -458,7 +460,22 @@ bool Aquamarine::CDRMAtomicImpl::commit(Hyprutils::Memory::CSharedPointer<SDRMCo
...
+    // Retry on wake only: enable modeset while the output was previously disabled.
+    const auto& STATE            = connector->output->state->state();
+    const bool  isDpmsOnWake     = !data.test && data.modeset && STATE.enabled && !connector->output->enabledState;
+    bool        ok               = request.commit(flags);
+    if (!ok && isDpmsOnWake) {
+        constexpr std::array<int, 3> BACKOFFS_MS = {100, 250, 500};
+        for (size_t i = 0; i < BACKOFFS_MS.size() && !ok; ++i) {
+            backend->log(AQ_LOG_DEBUG, std::format("atomic drm: DPMS ON commit failed, retry {}/{} in {}ms", i + 1, BACKOFFS_MS.size(), BACKOFFS_MS[i]));
+            std::this_thread::sleep_for(std::chrono::milliseconds(BACKOFFS_MS[i]));
+            ok = request.commit(flags);
```
vaxerski: absolutely not we will not hold the state ransom with blocking sleeps. You can add a timer that sends a .frame to the compositor in this case but a sleep() is unacceptable

dottorblaster: Will do, sorry for the caveman approach

dottorblaster:
It became way more invasive in the end. Feel free to drop it if you don't want to address this corner case.

vaxerski (commented):
I think a better way would be to extract timers into the core backend because headless has timers too. Then you could C++ify them like we do in hl

PR #25, opened 2024-07-24
displaylink fix: ignore primary argument when using evdi drivers

nktnet1:
Fixes #22.

A slightly more robust way to address the issue than the hack provided in #22 - done by checking if the driver being used for the given GPU is `evdi`. If so, set `primary` to `{}`.

Unsure of the following:
- whether this still falls under "hacky code" mentioned in Hyprland's [PR Guidelines](https://wiki.hyprland.org/Contributing-and-Debugging/PR-Guidelines/#pr-requirements)
- variables like `drmName` and `drmVer` are `auto` instead of `auto const` - have followed the same convention for the newly introduced `drmVerName` variable
- whether there exist use cases with `evdi` drivers and using the passed-in `primary_` argument (I assume no)

PR #279, opened 2026-04-16
drm: fix evdi outputs by adding CPU-copy fallback

### LINKED ISSUE #169: GBM/KMS issues with nvidia on dispaylink/evdi device
I have multiple systems with different hyperland versions. On the AMD system hyperland works with the DisplayLink based docking station. On the Intel + Nvidia laptop the rendering with hyperland fails. I built v0.48.1  and v0.8.0

I rebuilt Hyperland and aquamarine to produce these logs

[hyprland_displaylink.txt](https://github.com/user-attachments/files/20084968/hyprland_displaylink.txt)

sgtaziz:
evdi outputs never get a modeset on NVIDIA cards. Connector shows up and modes list, but drmModeAtomicCommit never runs. Reproducible on NVIDIA systems, using evdi's pyevdi `test_mode_changed_handler_called_after_connect` test, which passes in under 0.1s on KDE but times out at 60s here.
Supersedes #25, probably fixes #169.

evdi has no render device, so `CDRMRenderer::attempt` fails on it. The old workaround in `registerGPU` cleared primary for evdi so commits would skip the blit path and try a direct KMS import of the primary's buffer. That works for AMD/Intel (LINEAR dmabufs cross-import fine), but clearing primary also drops the multigpu LINEAR hint in the allocator, so on NVIDIA the primary buffer comes back `BLOCK_LINEAR_2D` and `drmModeAddFB2` on evdi refuses it.

Keeps primary set for evdi so output swapchains stay LINEAR, and adds a sink-only path. When CDRMRenderer::attempt fails on a secondary, mark it `sinkOnly`, leave `rendererState.renderer` null, and have `shouldBlit()` return false. Commits then fall into the existing direct-import branch. On AMD/Intel primaries that's enough.

On NVIDIA primaries `drmPrimeFDToHandle` on evdi fails because NVIDIA's dmabuf exporter refuses `dma_buf_attach` from foreign drivers. When that happens on a sink-only secondary, fall back to a CPU copy, allocate a dumb buffer on evdi via the existing `CDRMDumbAllocator`, have the primary renderer `glReadnPixels` into an intermediate, swizzle R to B for XRGB/ARGB destinations, write it into the dumb buffer's scanline, and build the FB from the dumb buffer. A `sinkCpuCopy` flag latches so subsequent commits skip the failed direct attempt.

I observed same last-resort method in other compositors as well, including KWin. The existing GBM-renderer mgpu blit path is untouched.

Tested on NVIDIA RTX 5090 + evdi with the test script bundled with pyevdi in the evdi repository.
Before: `AssertionError: Expected 'my_mode_changed_handler' to have been called` at 60.04s.
After: passes in under a 0.1s (0.04s average on my system).

### DIFF
--- a/include/aquamarine/backend/DRM.hpp
+++ b/include/aquamarine/backend/DRM.hpp
@@ -440,6 +440,8 @@ namespace Aquamarine {
         std::vector<SDRMFormat>                                       glFormats;
 
         Hyprutils::Memory::CSharedPointer<CDRMDumbAllocator>          dumbAllocator;
+        bool                                                          sinkOnly    = false;
+        bool                                                          sinkCpuCopy = false;
 
         bool                                                          atomic = false;
 
--- a/include/aquamarine/backend/DRM.hpp
+++ b/include/aquamarine/backend/DRM.hpp
@@ -498,6 +498,7 @@ namespace Aquamarine {
         uintptr_t                                                     nextPageFlipID();
 
         Hyprutils::Memory::CSharedPointer<CDRMDumbAllocator>          dumbAllocator;
+        bool                                                          cpuCopyFallback  = false;
 
         bool                                                          atomic = false;
 
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -1978,7 +1978,7 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
 
             drmFB = CDRMFB::create(NEWAQBUF, backend, nullptr); // will return attachment if present
         } else {
-            if (!(backend->sinkOnly && backend->sinkCpuCopy))
+            if (!backend->sinkOnly || !backend->sinkCpuCopy)
                 drmFB = CDRMFB::create(STATE.buffer, backend, nullptr); // will return attachment if present
 
             if (!drmFB && backend->sinkOnly && backend->dumbAllocator) {
@@ -1998,12 +1998,10 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
                     mgpu.swapchain = CSwapchain::create(backend->dumbAllocator, backend.lock());
 
                 auto bufDma  = STATE.buffer->dmabuf();
-                auto OPTIONS = mgpu.swapchain->currentOptions();
+                auto OPTIONS = swapchain->currentOptions();
                 OPTIONS.size = STATE.buffer->size;
                 if (OPTIONS.format == DRM_FORMAT_INVALID)
                     OPTIONS.format = bufDma.format;
-                if (OPTIONS.length == 0)
-                    OPTIONS.length = 2;
                 OPTIONS.multigpu = false;
                 OPTIONS.cursor   = false;
                 OPTIONS.scanout  = true;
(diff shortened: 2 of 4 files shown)


PlasmaPower:
I think this is duplicating the `CDRMRenderer::readBuffer` code and its existing usage in blit which already has a CPU-copy fallback path. I'm not sure what exact codepaths evdi takes, why wasn't the blit codepath working before and could it be fixed? Even if not, I think readBufferToDumb can be deduplicated with readBuffer.

sgtaziz:
Good points, a few things to note:

Couldn't use blit() directly because it runs GL on the secondary renderer's EGL context, and evdi has no secondary renderer. `CDRMRenderer::attempt` on its fd fails. The existing `intermediateBuf` fallback only covers the source side (primary reads to CPU, secondary uploads as texture, secondary still draws).

Tried the other angle: have the primary's renderer do the blit with the evdi dumb buffer as the destination FBO. `createEGLImage` on the dumb dmabuf actually succeeds on NVIDIA, but `glCheckFramebufferStatus` then fails with `GL_INVALID_OPERATION`. NVIDIA won't bind an evdi-exported dmabuf as renderbuffer storage. So the pure-GPU path isn't available on this combo.

Followed the dedup suggestion: folded the CPU fallback into blit() itself. Destination-side failures (`verifyDestinationDMABUF`, `createEGLImage`, or `glCheckFramebufferStatus`) now fall through to a CPU write if `to` has `BUFFER_CAPABILITY_DATAPTR`, reusing `readBuffer` for the readback. Also removed `readBufferToDumb` in favor of this. Let me know if that looks any better

sgtaziz:
Hi all, just wanted to follow up on this. DisplayLink/evdi currently does not work on Nvidia systems without a CPU-copy fallback. I welcome all feedback/changes if necessary!

sgtaziz:
Hi @PlasmaPower & @vaxerski 

Sorry to ping you guys, but just wanted to get some info on this. Unfortunately NVIDIA users still cant use displaylink/evdi displays currently at all on Hyprland

Just wanted to know if there is any interest in merging this PR so I know if I should try to resolve the merge conflicts, or if you guys want a different implementation altogether

Thank you both and all the maintainers for all your hard work

vaxerski:
does any other environment do this? CPU copying is abysmal in performance esp on buffers this size and I don't think this is a good idea

sgtaziz:
@vaxerski I do agree the performance isn't great, but its a last resort fallback that only triggers on the NVIDIA + evdi combo, where the alternative is evdi/displaylink displays just not working at all w/ NVIDIA

in terms of other environments:
- KWin: [`importWithCpu()`](https://github.com/KDE/kwin/blob/a586a4d2041435c83129a0e4e19eab9140645ec2/src/backends/drm/drm_egl_layer_surface.cpp#L572-L618), glReadnPixels straight into a dumb buffer on the sink. its the last of three import modes they try [in order](https://github.com/KDE/kwin/blob/a586a4d2041435c83129a0e4e19eab9140645ec2/src/backends/drm/drm_egl_layer_surface.cpp#L421-L437), and the gpu copy one [gives up if the sink has no render device](https://github.com/KDE/kwin/blob/a586a4d2041435c83129a0e4e19eab9140645ec2/src/backends/drm/drm_egl_layer_surface.cpp#L446-L452) which is exactly evdi
- Mutter: [`copy_shared_framebuffer_cpu()`](https://github.com/GNOME/mutter/blob/ea706ab9b328459e26833a718af2ae09ab45078d/src/backends/native/meta-onscreen-native.c#L1384-L1437), [called](https://github.com/GNOME/mutter/blob/ea706ab9b328459e26833a718af2ae09ab45078d/src/backends/native/meta-onscreen-native.c#L1488-L1492) once zero copy import and the primary gpu blit have both failed
- wlroots/smithay don't have it from what I found, so nvidia users can't use evdi/displaylink there either

it only applies to secondary kms devices and `blit()` takes the same path as before when the secondary renderer inits fine, so there should be no impact on other users

i can narrow it to `drmVerName == "evdi"` or put it behind a flag, i left it capability based since that's what KWin and Mutter do. can also cache the readback buffer instead of allocating per frame and read back as BGRA where supported so the XRGB path is a straight memcpy

but ill leave the trade offs to you, if you still don't think its worth ill close the PR to keep the noise down

vaxerski:
I would narrow it down to evdi yeah. But if kwin and g-ew-nome do it this way, that's fine. Just a conflict.

sgtaziz:
Done, narrowed it down to evdi. Had a conflict with #310 but kept it working, only had to not clear primary for evdi since its output swapchain is allocated against it and clearing it blanked the display. CPU copy is still last resort only, same idea as KWin and Mutter. Tested on a 5090 + evdi

aivars:
Thanks for this patch! With a locally adapted version on Aquamarine 0.15.0 / Hyprland 0.56.2, my Elgato Prompter now displays an image after restarting the session.
Previously it was detected but failed modesetting with drmPrimeFDToHandle failed.
  
Setup: Omarchy, RTX 5070, NVIDIA 610.57.04, EVDI 1.15.1, DisplayLink 6.3.
  
The adaptation includes commit-snapshot fence handling and disabling queued commits for EVDI. This confirms basic functionality; CPU usage and sustained performance haven’t been measured yet.

PR #281, opened 2026-04-17
refactor: avoid repeated vector access in backend and DRM loops

B2krobbery:
This removes a bit of repetition in a couple of loops by caching
implementations.at(i) and  crtcs.at(i) into local references.
No functional changes, just makes the code a bit cleaner and avoids
repeated  .at(i)  calls.
Touched:
-backend init loop in Backend.cpp
-CRTC assignment loop in DRM.cpp (recheckCRTCs)
Builds fine, logs look normal.


vaxerski (approved):
:+1:

PR #282, opened 2026-04-23
Enable libinput plugin loading via environment variable

flowln:
I've kept it disabled by default (when no environment variable is defined) so as to not change the default behavior, since it could have security implications. The implementation follows what is [documented in the API](https://wayland.freedesktop.org/libinput/doc/1.31.0/api/group__base.html#ga4d706c69b10714db54d4c8ae7343549c), particularly in relation to the code placement.

Since the 1.30.0 release of libinput is only a couple months old, I've added a version check so that we can continue supporting these older versions too, without this feature.


vaxerski (commented):
what's the practice around this on other environments? I can see a security problem with environment hijacking here, technically.

Not that something that runs as your user can't do other things already, though.

flowln:
> what's the practice around this on other environments? I can see a security problem with environment hijacking here, technically.
> 
> Not that something that runs as your user can't do other things already, though.

My motivation to make this change is so that I can load the button debounce plugin already inside libinput, because my keyboard sometimes sends extra key presses, which is annoying. There are a couple other built-in plugins too, though I don't know them in-depth (https://gitlab.freedesktop.org/libinput/libinput/-/blob/main/src/libinput-plugin.c?ref_type=heads#L508). They are probably used to handle some specific hardware scenarios too though.

Also, it seems that these plugins run in a sandboxed environment, though it's still in active development ([ref](https://gitlab.freedesktop.org/libinput/libinput/-/releases/1.31.1))

vaxerski:
that does not answer my question

PR #289, opened 2026-04-30
session: restore unconditional fallback to any render node

cornedor:
Restores the unconditional any-render-node fallback that existed before #235, which is needed for split-IP-block setups where the render device and KMS device have different udev parents (Apple Silicon being the most prominent example).

## Regression timeline

  | commit | change | effect on Apple Silicon |
  |---|---|---|
  | pre-#235 | match by PCI parent, fallback to first render node | works (PCI match fails → fallback opens renderD128) |
  | 35fa4a9 (#235) | match by generic udev parent, fallback removed | breaks (no parent match, no fallback → renderNodeFd = -1) |
  | 1dbbba6 (#237) | fallback re-added, gated by `!isStandardBus` | still broken (asahi & apple-drm are *platform* devices → gate blocks fallback) |
  | 8ed71bb | revert of #237 | unchanged (still broken) |
  | this PR | unconditional fallback (mirrors pre-#235 behavior) | fixed |

PR #237 only triggered the fallback when `cardDevice` had neither a PCI parent  nor a platform parent. The intent was "non-standard buses" (USB-attached EVDI,  virtual devices). But on Apple Silicon both `asahi` and `apple-drm` are platform  devices, so the gate blocked the fallback and #237 didn't actually fix #260, hence  the revert.

## Symptom on affected systems

  $ wayland-info | grep "main device"
          main device: 0xE202 (/dev/dri/card2)         # apple-drm, display-only

  $ # any GTK4 / EGL client
  libEGL warning: failed to get driver name for fd -1
  libEGL warning: MESA-LOADER: failed to retrieve device information
  GL_RENDERER:  llvmpipe (LLVM 22.1.1, 128 bits)        # software fallback

For example: https://github.com/ghostty-org/ghostty/discussions/12517

  After this patch:
  $ wayland-info | grep "main device"
          main device: 0xE201 (/dev/dri/renderD128)    # asahi GPU

 $ ./repro-test-app   # GTK4 + GL 4.3 + SSBO at binding 1, mirrors Ghostty's cell_bg pass
  GL_RENDERER:  Apple M2 Max (G14C B1)
  GL_VERSION:   4.6 (Core Profile) Mesa 26.0.5

Ghostty's per-cell background colors render correctly now.

### DIFF
--- a/src/backend/Session.cpp
+++ b/src/backend/Session.cpp
@@ -1,5 +1,8 @@
 #include <aquamarine/backend/Backend.hpp>
+#include <cerrno>
+#include <cstdlib>
 #include <fcntl.h>
+#include <vector>
 
 extern "C" {
 #include <libseat.h>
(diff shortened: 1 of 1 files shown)


uorbe001:
I built hyprland (master) with this patch, and I can confirm the performance issues on the asahi M1 are gone, so this should resolve #290.

jwbron:
Confirming this fixes the symptom on M1 Pro / Fedora Asahi 43 too (`main_device` flips from `0xE202` to `0xE280`, hardware WebRender returns, YouTube fullscreen native). Filed #290 with the full diagnosis just before noticing this PR existed.

One concern about the unconditional shape: it is functionally identical to the reverted #237 minus the `!isStandardBus` gate, and #237 was reverted per #260 because the unconditional fallback regressed multi-renderD setups (NixOS user's hyprpaper SIGABRT on `eglQueryDeviceStringEXT EGL_BAD_PARAMETER`). The most plausible failure mode is the fallback opening a non-renderable renderD on systems with more than one (eGPU + iGPU, evdi, capture devices). Picking the first one and handing it to EGL is what triggers the BAD_PARAMETER.

Counting renderD nodes first and only falling back when there is exactly one keeps Asahi (single renderD128) working while leaving multi-renderD systems on the existing `gpu->fd` behavior, so they fail loud rather than silently picking the wrong renderer. I posted that variant as #291 in case it is useful as a follow-up. This PR's simpler shape is preferable as the baseline if maintainers are confident #260's failure mode was unrelated to the unconditional fallback. Guidance from a maintainer on what specifically broke under #237 would help judge whether the count guard is needed.

cornedor:
Thanks for the response. 

Rather than ship the count-guard as a band-aid, I replaced the unconditional fallback with a drmGetDevices2 + bus-type-scoped approach in f0341f0. Four stages, first hit wins:

  1. AQ_RENDER_NODE env var (new; best-effort path override).
  2. drmGetDevices2 kernel pairing, matching cardX by st_rdev against DRM_NODE_PRIMARY.
  3. Fallback only when cardBusType is DRM_BUS_PLATFORM, matching against another DRM_BUS_PLATFORM device with a render node.
  4. Give up; renderNodeFd stays -1.

The safety property: stage 3 cannot fire on a DRM_BUS_PCI cardX. Multi-renderD PCI systems (eGPU + iGPU, capture cards, evdi) physically cannot regress with a silently wrong renderer. They either succeed via stage 2 kernel-authoritative pairing or fail loudly.

This is the same algorithm KWin uses in src/core/gpumanager.cpp::findCompatibleRenderDevice.

Verified on M1 Pro Asahi: stage 3 fires as designed, /dev/dri/renderD128 opens, hardware acceleration works, no EGL_BAD_PARAMETER in the log. The AQ_RENDER_NODE knob is a bonus escape hatch for any future weird hardware (PCI passthrough, exotic eGPU layouts) without needing more code changes.

apilot:
Thanks for this PR! I've been testing it on a multi-GPU + DisplayLink setup and found a gap in Stage 3 that leaves EVDI devices without a render node.
My system:
- Intel i915 (card0, renderD128, DRM_BUS_PCI)
- NVIDIA (card1, renderD129, DRM_BUS_PCI)
- EVDI DisplayLink (card2, no render node, DRM_BUS_PLATFORM)
What happens with current Stage 3:
Stage 3 only matches DRM_BUS_PLATFORM render nodes. On my system, both real GPUs are DRM_BUS_PCI, so there are zero platform-bus render nodes to find. EVDI gets no render node → CDRMRenderer::attempt falls back to EVDI card fd → EGL fails with EGL_BAD_PARAMETER.
This would also affect any DisplayLink/EVDI user with a standard dual-GPU laptop (Intel + NVIDIA) — a very common configuration.
Fix — add Stage 4: PCI render node fallback (DRM_BUS_PLATFORM only)
After Stage 3 fails for DRM_BUS_PLATFORM devices, I added a fallback to PCI render nodes with a 2-pass approach:
- Pass 0: Skip devices whose PCI driver symlink (/sys/bus/pci/devices/{domain:bus:dev.func}/driver) contains "nvidia". This prefers Mesa/Intel render nodes which are compatible with EVDI's GBM expectations.
- Pass 1: Accept any PCI render node (NVIDIA as last resort).
The NVIDIA skip is necessary because NVIDIA EGL + GBM buffers are incompatible with EVDI Mesa/GBM — I tested with renderD129 and got vertical stripes on the DisplayLink output.
Driver detection (avoids hardcoding):
auto busId = std::format("{:04x}:{:02x}:{:02x}.{:x}",
    d->businfo.pci->domain, d->businfo.pci->bus,
    d->businfo.pci->dev, d->businfo.pci->func);
char driverBuf[256];
ssize_t len = readlink(
    std::format("/sys/bus/pci/devices/{}/driver", busId).c_str(),
    driverBuf, sizeof(driverBuf) - 1);
// Check if driverBuf contains "nvidia"
Results with the additional stage:
- 3 monitors working: eDP-1 (Intel), HDMI-A-1 (NVIDIA), DVI-I-1 (DisplayLink via renderD128)
- 0 renderer errors (was 1945+ per session)
- Power cycle reconnects correctly, restores workspace
Happy to share my full patch if useful. Want me to open a separate PR or contribute the Stage 4 additions to this one?

jwbron:
Note, #291 was merged, so this PR is now redundant.

PR #291, opened 2026-04-30
drm: fall back to first render node on single-renderD systems when parent-syspath match fails

### LINKED ISSUE #290: M-series Asahi: parent-syspath render-node match fails, renderer falls back to KMS fd (software rendering)
## Summary

On M-series Apple Silicon (Asahi Linux), aquamarine 0.11.0 attaches its renderer to `/dev/dri/card2` (apple-drm KMS, no userspace GL/EGL driver) instead of the asahi GPU's render node `/dev/dri/renderD128`. Mesa-EGL clients fall back to llvmpipe: Firefox WebRender drops to `(Software)`, YouTube fullscreen judders, Slack scroll is choppy. `zwp_linux_dmabuf_v1.main_device` advertises `0xE202` (= card2) instead of `0xE280` (= renderD128).

The same Hyprland 0.54.3 + aquamarine 0.11.0 binary works correctly on a different Asahi host (M2-class, different DT layout), confirming this is a device-pick issue, not a Mesa or Hyprland bug.

## Repro

- Apple M1 Pro (`apple,t6000`)
- Fedora Asahi Remix 43, kernel 6.19.13 asahi
- aquamarine 0.11.0, Hyprland 0.54.3
- Mesa 25.3.6

DRM topology:

```
card1: asahi (render-only, no KMS), parent: platform/soc/406400000.gpu
card2: apple-drm (KMS-only, no GL),  parent: platform/soc/soc:display-subsystem
renderD128: asahi GPU's render node
```

## Diagnosis

`Session.cpp::resolveMatchingRenderNode` matches a render node to the KMS device by parent-syspath. On this host the asahi GPU and apple-drm KMS device sit under different parent platform devices, so the match fails, `renderNodeFd` stays `-1`, and `DRM.cpp:622` attaches the renderer with `gpu->fd` (the apple-drm KMS fd, no GL driver):

```cpp
rendererState.renderer = CDRMRenderer::attempt(backend.lock(),
    gpu->renderNodeFd >= 0 ? gpu->renderNodeFd : gpu->fd);
```

Aquamarine startup log:

```
DEBUG aquamarine: drm: Enumerated device .../406400000.gpu/drm/card1
DEBUG aquamarine: libseat: Device /dev/dri/card1 does not support kms
ERR   aquamarine: drm: Skipping device .../card1, not a KMS device
DEBUG aquamarine: drm: Enumerated device .../soc:display-subsystem/drm/card2
DEBUG aquamarine: libseat: Device /dev/dri/card2 supports kms
[no "matching render node" log: renderNodeFd stays -1]
```

## What does NOT work

- **`AQ_DRM_DEVICES=cardN`** cannot include card1 (asahi): aquamarine pre-filters non-KMS devices via `openIfKMS()` before applying the explicit list. The list then logs `Explicit device /dev/dri/card1 not found`.
- **`AQ_DRM_DEVICES=/dev/dri/by-path/...`** mangles paths containing `:` (apple-drm's stable by-path is `platform-soc:display-subsystem-card`). The list parser splits on `:` first, then canonicalizes each fragment.

## What does work

Per @tempergate in #260, commit `2455556b426fd82e119e4c98c57fcd7a54625478` (immediately before #235 generalized parent matching) "works correctly" on M-series Asahi. That commit had PCI-specific match plus an unconditional first-renderD fallback. PCI match always failed on Asahi (no PCI parent), so the fallback ran and picked renderD128.

#237 tried to bring the fallback back, gated on `!isStandardBus` (standard = PCI || platform). Asahi is on platform bus, so the gate excluded us; that patch never could have helped here. #237 was then reverted by `8ed71bb` per #260 ("breaks M1/M2, potentially more split-node drivers").

## Proposed fix

Re-add a first-renderD fallback to `resolveMatchingRenderNode`, but only when the system has exactly one renderD node. Counting first makes Asahi succeed (single renderD128) while leaving multi-renderD systems on upstream's existing `gpu->fd`-as-renderer behavior. Multi-renderD configs (eGPU + iGPU, evdi, capture devices) cannot be disambiguated from this layer without more info, and picking the wrong one is plausibly what triggered #260's hyprpaper SIGABRT on `eglQueryDeviceStringEXT EGL_BAD_PARAMETER`. Counting avoids that class of regression.

Draft PR to follow.

I would appreciate a pointer from a maintainer to the specific failure mode that triggered the #260 revert. #237 did not run for platform-bus systems, so the M1/M2 break may have come from a separate change in the same window (e.g., the evdi-skip removal in #235's `scanGPUs`); confirming that would help land a robust fix.

## Workaround in the meantime

Building this patch locally on Fedora Asahi with a versionlocked source-build over `/usr/lib64/libaquamarine.so.0.11.0`. Confirmed: `main_device` flips to `0xE280`, Firefox WebRender goes hardware, YouTube fullscreen is native.

jwbron:
Re-adds the pre-#235 first-renderD fallback in `Session.cpp::resolveMatchingRenderNode`, scoped to systems with exactly one renderD node. Fixes #290.

## Why no bus gate (vs #237)

#237 was gated on `!isStandardBus` where standard = PCI || platform. Asahi GPUs are on the platform bus, so the gate excluded the M-series users who actually need this fallback. Removing the gate makes the fallback unconditional in principle.

## Why count then fall back (vs #237's revert)

#237 was reverted per #260 ("breaks M1/M2, potentially more split-node drivers"). The most plausible cause of the regression is the unconditional fallback opening a non-renderable renderD on multi-renderD setups (eGPU + iGPU, evdi, capture devices); that matches the symptom in #260's hyprpaper SIGABRT on `eglQueryDeviceStringEXT EGL_BAD_PARAMETER`.

Counting renderD nodes first and falling back only when there is exactly one means multi-renderD systems revert to upstream's current `gpu->fd`-as-renderer behavior. Single-renderD setups (Asahi being the canonical case) get the fix; ambiguous setups fail loud rather than guessing wrong.

## Tested

Apple M1 Pro / Fedora Asahi Remix 43 / Mesa 25.3.6:

- `wayland-info`: `zwp_linux_dmabuf_v1.main_device` flips from `0xE202` (apple-drm KMS, no GL) to `0xE280` (asahi `renderD128`).
- Firefox `about:support` Compositing: `WebRender` (no `(Software)`).
- YouTube fullscreen: native.
- Slack scroll: native.
- Aquamarine startup log: `drm: No matching render node for /sys/devices/platform/soc/soc:display-subsystem, falling back to sole renderD on the system: /dev/dri/renderD128`.

## Open question

Marked draft. Maintainer guidance appreciated on the specific failure mode that triggered the #260 revert. #237 did not actually run for platform-bus systems, so the M1/M2 break may have been from a separate change in the same window (e.g., the evdi-skip removal in #235's `scanGPUs`); confirming that would help judge whether this single-renderD scoping fully addresses the concern.

## Out of scope

- `AQ_DRM_DEVICES` `:` separator escaping (so by-path paths with colons parse correctly). Separate concern; not blocking this PR.

### DIFF
--- a/src/backend/Session.cpp
+++ b/src/backend/Session.cpp
@@ -205,6 +205,69 @@ void Aquamarine::CSessionDevice::resolveMatchingRenderNode(udev_device* cardDevi
         udev_device_unref(dev);
     }
 
+    // Re-add the pre-35fa4a9 first-renderD fallback for the case where
+    // parent-syspath matching fails. Required on split-node platforms
+    // like Apple Silicon, where the GPU and KMS device sit under
+    // different parent platform devices and the match never succeeds.
+    // Without this, DRM.cpp's renderer attach falls back to gpu->fd
+    // (the KMS device fd); on apple-drm that fd has no GL/EGL driver,
+    // so Mesa is forced to llvmpipe and the desktop runs in software.
+    //
+    // We only fall back when there is exactly one renderD on the
+    // system. Multi-renderD configs (eGPU + iGPU, evdi, capture
+    // devices) can't be disambiguated from this layer; picking the
+    // wrong one is what regressed PR #237 per #260 (hyprpaper SIGABRT
+    // on eglQueryDeviceStringEXT). Counting first lets Asahi succeed
+    // while leaving multi-renderD systems unchanged from upstream.
+    //
+    // Refs hyprwm/aquamarine#260 (M1/M2 reports), PR #237 (unconditional
+    // for non-standard buses, reverted by 8ed71bb), commit 2455556b
+    // (last known-good on M-series Asahi per the #260 thread).
+    if (renderNodeFd < 0) {
+        int renderDCount = 0;
+        udev_list_entry* countEntry = nullptr;
+        udev_list_entry_foreach(countEntry, devices) {
+            const auto* path = udev_list_entry_get_name(countEntry);
+            auto        dev  = udev_device_new_from_syspath(session->udevHandle, path);
+            if (!dev)
+                continue;
+            const auto* devnode = udev_device_get_devnode(dev);
+            const auto* devtype = udev_device_get_devtype(dev);
+            if (devnode && devtype && strcmp(devtype, "drm_minor") == 0 && strstr(devnode, "renderD"))
+                ++renderDCount;
+            udev_device_unref(dev);
+        }
+
+        if (renderDCount == 1) {
+            udev_list_entry* fbEntry = nullptr;
+            udev_list_entry_foreach(fbEntry, devices) {
+                const auto* fbPath = udev_list_entry_get_name(fbEntry);
+                auto        fbDev  = udev_device_new_from_syspath(session->udevHandle, fbPath);
+                if (!fbDev)
+                    continue;
+                const auto* fbDevnode = udev_device_get_devnode(fbDev);
+                const auto* fbDevtype = udev_device_get_devtype(fbDev);
+                if (!fbDevnode || !fbDevtype || strcmp(fbDevtype, "drm_minor") != 0 || !strstr(fbDevnode, "renderD")) {
+                    udev_device_unref(fbDev);
+                    continue;
+                }
+                renderNodeFd = open(fbDevnode, O_RDWR | O_CLOEXEC);
+                if (renderNodeFd >= 0) {
+                    session->backend->log(AQ_LOG_WARNING,
+                        std::format("drm: No matching render node for {}, falling back to sole renderD on the system: {}",
+                            parentSyspath ? parentSyspath : "(unknown)", fbDevnode));
+                    udev_device_unref(fbDev);
+                    break;
+                }
+                udev_device_unref(fbDev);
+            }
+        } else if (renderDCount > 1) {
+            session->backend->log(AQ_LOG_WARNING,
+                std::format("drm: No matching render node for {}, refusing to guess among {} renderD candidates; leaving renderer to fall through to KMS fd (may force software rendering; set AQ_DRM_DEVICES to disambiguate)",
+                    parentSyspath ? parentSyspath : "(unknown)", renderDCount));
+        }
+    }
+
     udev_enumerate_unref(enumerate);
 }
 


uorbe001:
I've just built and tested this, and I can confirm this does fix the performance issues on asahi M1 too, so this and #289 both work, I suppose a maintainer should pitch in and decide which one is preferable? Thank you for these fixes btw.

tempergate:
> I've just built and tested this, and I can confirm this does fix the performance issues on asahi M1 too, so this and #289 both work, I suppose a maintainer should pitch in and decide which one is preferable? Thank you for these fixes btw.

I'm confirming the same, using asahi kernel under NixOS

ywaf:
build and tested, patch works well. Thank you :)

"m1 pro" macbook pro fc43.

vaxerski (commented):
generally looks good. I can't test or verify as I don't use fruity hardware. I don't see how this could pose a real problem either

vaxerski:
not merging as draft

jwbron:
Marked this as ready. I've been running this for about a week now with no issues.

vaxerski (approved):

waltmck:
I still see [this problem](https://github.com/hyprwm/Hyprland/discussions/11398) with OpenGL not working on the following Hyprland version:
```
Hyprland 0.55.1 built from branch v0.55.1-b at commit a47147bc095e5b3be3eb8bd04f0ac242b968cd4d clean ([gha] Nix: update inputs).
Date: 2026-05-13
Tag: v0.55.1, commits: -1

Libraries:
Hyprgraphics: built against 0.5.1, system has unknown
Hyprutils: built against 0.13.1, system has unknown
Hyprcursor: built against 0.1.13, system has unknown
Hyprlang: built against 0.6.8, system has unknown
Aquamarine: built against 0.11.0, system has unknown

Version ABI string: a47147bc095e5b3be3eb8bd04f0ac242b968cd4d_aq_0.11_hu_0.13_hg_0.5_hc_0.1_hlg_0.6
flags set:
nix
```

The date on that commit is after the date this was merged. Either Hyprland isn't using a recent enough aquamarine, or this commit did not fix that issue.

Edit: this issue persists even when using the latest version of aquamarine with Nix flake inputs:
```
inputs.hyprland = {
  url = "git+https://github.com/hyprwm/Hyprland?submodules=1";
  inputs.nixpkgs.follows = "nixpkgs";
  inputs.aquamarine.follows = "aquamarine";
};

inputs.aquamarine = {
  url = "github:hyprwm/aquamarine";
  inputs.nixpkgs.follows = "nixpkgs";
};
```

PR #296, opened 2026-05-09
backend: avoid duplicate auto-named outputs

### LINKED ISSUE #185: `hyprctl output` allows creation of duplicate output names
### Regression:
No
### Version:
```
Hyprland 0.49.0 built from branch main at commit 6a5f4f5954a64bac718e3938f062d045256e7aeb  (Nix: fix overlay application).
Date: Sat Jul 5 13:18:38 2025
Tag: v0.49.0-145-g6a5f4f59, commits: 6243
built against:
 aquamarine 0.8.0
 hyprlang 0.6.3
 hyprutils 0.8.0
 hyprcursor 0.1.12
 hyprgraphics 0.1.4


flags set:
debug
```

### Description:
if you create an output with the name that will be auto generated the next time, and then create another output with an auto generated name, you get two outputs with the same name, which shouldn't happen according to wayland spec

- the check in [HyprCtl.cpp](https://github.com/hyprwm/Hyprland/blob/main/src/debug/HyprCtl.cpp#L1460-L1461) is skipped because no name is passed when creating the second output
- [backend/Wayland.cpp](https://github.com/hyprwm/aquamarine/blob/main/src/backend/Wayland.cpp#L143) does not catch this
- all of this also applies to headless outputs
- i didn't experience a crash or anything, it is just annoying
- sry if i should have created this issue in the Hyprland repo, i just felt like the relevant check should be added to aquamarine so that is where i created the issue

Image

### Steps to reproduce:
```bash
# start new Hyprland instance on wayland backend
# make sure there is exactly one output named WAYLAND-1

hyprctl output create wayland WAYLAND-2
hyprctl output create wayland # this creates another output called WAYLAND-2

hyprctl monitors # see that you have two outputs with the same name
```

UncleJ4ck:
Both the wayland and headless backends generated `WAYLAND-N` / `HEADLESS-N`
auto names from a monotonic counter without checking for existing
outputs. If a user creates an output with an explicit name that matches
a future counter slot (e.g. `hyprctl output create wayland WAYLAND-2`),
the next auto-named output collides on the same name.

Skip past any taken slot when generating an auto name, and reject
explicit creates whose name is already in use.

Fixes #185.

### DIFF
--- a/src/backend/Headless.cpp
+++ b/src/backend/Headless.cpp
@@ -175,7 +175,19 @@ std::vector<SDRMFormat> Aquamarine::CHeadlessBackend::getCursorFormats() {
 }
 
 bool Aquamarine::CHeadlessBackend::createOutput(const std::string& name) {
-    auto output = SP<CHeadlessOutput>(new CHeadlessOutput(name.empty() ? std::format("HEADLESS-{}", ++outputIDCounter) : name, self.lock()));
+    std::string outputName = name;
+    if (outputName.empty()) {
+        // skip past any auto-name slots already claimed by an explicit createOutput,
+        // so we never hand out a duplicate HEADLESS-N (see #185).
+        do {
+            outputName = std::format("HEADLESS-{}", ++outputIDCounter);
+        } while (std::ranges::any_of(outputs, [&outputName](const auto& o) { return o->name == outputName; }));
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
even though I'd classify this as a user-problem

PR #297, opened 2026-05-09
drm: clear stale color state on modeset

### LINKED ISSUE #127: DRM properties set by other compositors are not reset and mess with AQ
Some (all?) DRM properties keep their values after the compositor exits. Not all of such properties are reset when HL starts, some of those properties aren't even supported by AQ. This might result in a picture difference if HL is started after some other compositor. E.g. kwin HDR and CM settings affect HL's gamma and saturation:

1. start kwin
2. enable hdr and adjust some settings
3. exit kwin
4. start HL.
HL renders in SDR mode with CM settings suitable for HDR mode which results in washed out colors.

Should be some generic way to reset everything. https://dri.freedesktop.org/docs/drm/gpu/drm-kms.html#c.drm_mode_config_reset?

UncleJ4ck:
A previous compositor (kwin etc.) can leave HDR, gamma, degamma and CTM
properties set on a connector when it exits. Aquamarine only re-emitted
these on user commits, so anything we don't actively manage stayed at
the previous compositor's value, manifesting as washed-out colors or
shifted saturation when starting Hyprland after kwin.

HDR (#250) and CTM-with-non-identity (#256) were already covered. Extend
the modeset path to also:

- re-send `gamma_lut` / `degamma_lut` on modeset; an empty `STATE.gammaLut`
  produces a zero blob, which clears the kernel state. only ride this
  path when the prop actually exists, otherwise we'd log a spurious
  `no gamma_lut prop` error on every modeset for connectors without
  programmable gamma.
- re-send CTM unconditionally on modeset, identity included, so the
  kernel always replaces a stale matrix with our current one. the blob
  builder produces a real identity matrix when `STATE.ctm == Mat3x3()`,
  so we never send `blob_id=0` (which #256 documented as problematic).

Fixes #127.

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -2126,7 +2126,12 @@ bool Aquamarine::CDRMOutput::commitState(bool onlyTest) {
     else
         data.calculateMode(connector);
 
-    if ((data.modeset && STATE.ctm != Mat3x3()) || (COMMITTED & COutputState::eOutputStateProperties::AQ_OUTPUT_STATE_CTM))
+    // always re-send CTM on modeset, identity included, otherwise a stale matrix
+    // left by a previous compositor (e.g. kwin's color management) survives our
+    // session. prepareConnector builds a real identity blob when STATE.ctm is
+    // identity, so we never send blob_id=0 (which some drivers mishandle, see #256).
+    // (see #127)
+    if (data.modeset || (COMMITTED & COutputState::eOutputStateProperties::AQ_OUTPUT_STATE_CTM))
         data.ctm = STATE.ctm;
 
     bool ok = connector->commitState(data);
--- a/src/backend/drm/impl/Atomic.cpp
+++ b/src/backend/drm/impl/Atomic.cpp
@@ -369,10 +369,15 @@ bool Aquamarine::CDRMAtomicImpl::prepareConnector(Hyprutils::Memory::CSharedPoin
         return false;
     };
 
-    if (STATE.committed & COutputState::AQ_OUTPUT_STATE_GAMMA_LUT)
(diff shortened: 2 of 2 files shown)


vaxerski (approved):

cushycush:
Hey, a heads up that I think it might've introduced a black-screen on some setups, wanted to flag it in case I'm reading the path wrong.

I'm running a Hyprland 0.55-based fork with `render:cm_enabled=false` on Intel UHD 770 / i915. Under atomic mode, every panel goes black after the first modeset. Legacy DRM works fine. The framebuffer itself is correct: grim reads back the rendered desktop. But `modetest -M i915 -p` shows the CRTC's CTM property is all-zeros after the first modeset, so the kernel multiplies every output pixel by zero.

The PR's commit message says the blob builder produces identity when `STATE.ctm == Mat3x3()`, but I don't think it does. `Hyprutils::Math::Mat3x3()` zero-initializes its 9 floats, and the builder at `impl/Atomic.cpp:419-423` runs the matrix through `doubleToS3132Fixed` and writes it verbatim. So a default-constructed Mat3x3 lands in the kernel as a zero blob, not identity. Hyprland's `setCTM` only fires from the cm_enabled=true renderer path, so forks running with it disabled never overwrite the zero default.

Easy fix would be treating a default-constructed `Mat3x3` as identity inside the blob builder, contained to this file. Flipping `Hyprutils::Math::Mat3x3()`'s default ctor to identity upstream would also close it but has broader blast radius for anyone quietly relying on zero. Anyone hitting this before an upstream fix can seed `setCTM(identity())` at output creation as a workaround (no-op under legacy, seeds an identity matrix under atomic instead of all-zeros). Happy to send a PR if it helps, but figured I'd flag and let you take it since you're already in this file. Am I missing why the default ctor would produce identity?

PR #299, opened 2026-05-09
treewide: move Math.{hpp,cpp} from backend/drm to utils

UncleJ4ck:
The matrix helpers in `src/backend/drm/Math.{hpp,cpp}` are not DRM-specific.
The header had a long-standing FIXME asking to move them under utils where
generic helpers live alongside `FormatUtils` and `Shared`.

Move `src/backend/drm/Math.cpp` -> `src/utils/Math.cpp` and
`src/backend/drm/Math.hpp` -> `src/include/Math.hpp`. `Renderer.cpp`'s
`#include "Math.hpp"` still resolves through the existing `src/include`
private include path, so no other source change is needed. The
`CMakeLists.txt` glob picks the new locations up automatically.

Drop the FIXME now that it has been done.


vaxerski (commented):
I meant hyprutils xD

UncleJ4ck:
closing, this is the wrong shape per your review. The right cleanup is migrating callers to `Hyprutils::Math::Mat3x3` and deleting `Math.{cpp,hpp}`, not moving the file. Hit a snag doing that properly: aquamarine's `matrixProjection` negates t[3] and t[4] to bake in the KMS Y-flip, while `Mat3x3::outputProjection` doesn't, so a naive migration would invert Y on the multi-GPU blit path. Will revisit as a follow-up.

PR #305, opened 2026-05-16
Overlay plane

UjinT34:
Most of the HW have at least 3 planes per crtc: primary, cursor and overlay. On nvidia it's max 3 planes. Not sure about other HW. This PR focuses on adding support for an overlay plane. It'll be used to avoid leaving DS when a notification/special workspace or something similar shows up above fullscreen window.
May include some extra features.

- [x] track extra planes states
- [x] commit extra planes states
- [ ] hl overlay PR https://github.com/hyprwm/Hyprland/pull/14655
- [ ] reset unknown properties 
- [ ] ~~handle drm color pipeline and vendor-specific CM props~~
- [ ] ~~provide an api to hide vendor-specific CM details and drm color pipeline implementation differences~~
- [ ] ~~hl CM PR~~

Includes #96

### DIFF
--- a/include/aquamarine/backend/DRM.hpp
+++ b/include/aquamarine/backend/DRM.hpp
@@ -4,7 +4,9 @@
 #include "../allocator/Swapchain.hpp"
 #include "../output/Output.hpp"
 #include "../input/Input.hpp"
+#include <cstdint>
 #include <hyprutils/memory/WeakPtr.hpp>
+#include <vector>
 #include <wayland-client.h>
 #include <xf86drmMode.h>
 #include <optional>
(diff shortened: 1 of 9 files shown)


fxzzi:
is cursor plane possible too? people have issues with the kde login thing persisting their cursor on screen because Hyprland doesn't clear the cursor plane (i think this is the diagnosis? not sure, i don't use the kde login manager)

UjinT34:
hw cursor uses cursor plane

fxzzi:
ok cool then idk wtf sddm is doing

UjinT34:
Should be usable. Lacks error handling, errors are ignored.

fxzzi:
set render:use_overlay_plane to 1, looks like the overlay plane is being used just based off the fact there is no blur and its a lower framerate

but when in ds and opening an overlay the overlay is just completely invisible

when in hdr with ds OFF, the overlay planes have blown out colours

UjinT34:
AQ side is usable. HL needs work. DS returns early and doesn't get to overlay stuff. Overlay stuff itself doesn't go through CM. Currently it just avoids black screens and CM changes when some notification appears during gameplay.

fxzzi:
Ah ok, well in that case it's working fine, notifications don't show and HDR continues to work

PR #310, opened 2026-05-25
Fix DisplayLink/evdi monitor hotplug

mwz:
DisplayLink/evdi devices provide a KMS display device, but they do not provide a usable EGL renderer. Aquamarine already treats `evdi` as display-only, but after the secondary renderer lifetime changes, hotplugged connectors try to initialise a renderer before the output is announced. Since renderer creation failed for `evdi`, the connect path returned early and the evdi monitors never showed up.

The key logs that pointed to this were the following:

```text
CDRMRenderer(drm): drmGetDevice failed
CDRMRenderer(drm): Can't create renderer, no matching devices found
drm: initMgpu: no renderer
drm: Failed to update renderer state for DVI-I-* on connect
```

This change marks evdi backends as not needing a renderer, skips the startup renderer probe for them, and lets `initMgpu()` succeed for these display-only devices. Normal DRM GPUs still require renderer initialisation, but DisplayLink/evdi outputs can now finish connecting and be announced when hotplugged.

This fixes for me the regression, which was introduced in 0.11.

### DIFF
--- a/include/aquamarine/backend/DRM.hpp
+++ b/include/aquamarine/backend/DRM.hpp
@@ -450,6 +450,8 @@ namespace Aquamarine {
             Hyprutils::Memory::CSharedPointer<CDRMRenderer> renderer; // may be null if creation fails
         } rendererState;
 
+        bool                                                          rendererRequired = true;
+
         Hyprutils::Memory::CWeakPointer<CBackend>                     backend;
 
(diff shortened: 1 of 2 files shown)


TheFissk:
I was experiencing the same regression and this works fo rme as well.

vaxerski (approved):
tnx

Issue #308, opened 2026-05-21
Lid-resume eDP panel wedge persists on Intel i915 with f5cdaa8 in place (no memory pressure)


johnkferguson:
## Summary

After upgrading from aquamarine 0.10.x (Hyprland 0.54) to a post-v0.11.0 build that includes [`f5cdaa8`](https://github.com/hyprwm/aquamarine/commit/f5cdaa8) ("drm: CRTC starvation recovery + clear stale page-flip state after suspend (#254)"), the eDP panel still wedges on a lid-close → suspend → lid-open cycle. Same wedge symptom as #304 (backlight on, no scanout, only compositor restart recovers), but the f5cdaa8 recovery does not catch this case. The wedge is **not** memory-pressure related: a per-process RSS poller running every 10s shows Hyprland holding ~50 MB across the entire reproduction.

Filing because f5cdaa8 was the targeted fix for this family and it's clearly only partial on Intel i915 setups. Cross-reference #304 (same wedge symptom, AMD, idle-DPMS trigger; this report is i915 + lid-suspend trigger).

## System

- **GPU:** Intel Raptor Lake-U [8086:a7ac], i915 driver. Single eDP-1 internal panel, no external outputs.
- **Hyprland:** 0.55.1 (nixpkgs build, commit `a47147bc`), Lua config.
- **Aquamarine:** built from main at `ab2b0af6` (10 commits past v0.11.0). f5cdaa8 verified active in the loaded library — all eight of its new log strings (`Clearing stale page-flip state for ...`, `backup slot {} crtc {} assigned to disabled {}`, `Connector {} enabledState changed`, etc.) are present in the running `libaquamarine.so`.
- **Kernel:** 6.18.31 (NixOS unstable).
- **Session manager:** UWSM with atomic DRM (no `AQ_NO_ATOMIC`).
- **Other libs:** Hyprutils 0.13.1, Hyprgraphics 0.5.1, Hyprcursor 0.1.13, Hyprlang 0.6.8.

## Reproduction

Plain lid-close to lid-open cycle. Both short (~6 min) and long (~7 hour) suspends reproduce the wedge. The full lifecycle:

1. Compositor running normally, Hyprland RSS ~50 MB.
2. User closes lid → logind suspends (s2idle).
3. After N minutes, user opens lid → kernel resumes.
4. Lock screen (hyprlock, started on lid-close via Hyprland's `switch:on:Lid Switch` binding) flashes once — clock + password entry briefly visible.
[...]

Issue #306, opened 2026-05-17
Source 1 games regression: Low fps, frozen/broken direct scanout, invisible cursor


Szwagi:
Edit: Seems like all Source 1 games are affected. Posted more games that are more relevant and easier to launch in a comment.

CS:GO (not CS2) had a slightly buggy cursor before, but that was the only issue.
If the cursor was not on top of any UI element it'd flicker and therefore also flicker direct scanout. I assume this is a quirk of the game not Hyprland/Aquamarine.

After a recent commit I see these issues:
1. FPS drops significantly (locked 800 fps drops to 100-200 fps).
When that happens direct scanout reports as disabled.
Sometimes fixes itself after disabling and enabling fullscreen or changing workspace.
2. Sometimes the screen freezes when the cursor would appear (like opening the in-game chat or console).
Fixes itself after changing workspace.
3. The cursor is invisible. Both software and hardware cursors.

None of this happens with direct scanout disabled in the config or tearing disabled.

Tested on Hyprland 0.55.2.

Last version that worked correctly is https://github.com/hyprwm/aquamarine/commit/813c1e8981893c11e118b19c125d6bc282f51765.
If you apply only https://github.com/hyprwm/aquamarine/commit/be35f75ac305f430f5f9d89b5f5a4af59ca7567e on top of it, it's broken.
It's also broken if you apply all the commits in between. Not applying the broken commit to these results in a black screen.

To recreate:
1. NVIDIA (Unsure if relevant here)

1. Enable direct scanout and tearing

2.
```lua
hl.window_rule({
    name  = "csgo",
    match = { 
        class = "csgo_linux64",
    },
    fullscreen = true,
    content = "game",
    suppress_event = "fullscreen maximize activate activatefocus",
    no_vrr = true,
    immediate = true,
})
```

3. Launch the game
- <https://store.steampowered.com/app/4465480/CounterStrikeGlobal_Offensive/>
- Install libSDL2 (for Wayland support, the one Valve ships is ANCIENT and broken)
- `LD_PRELOAD="/usr/lib/libSDL2.so" %command% -vulkan -console`

4. You can toggle the console with ~ key
- If the cursor is invisible then it's broken and all the issues mentioned should be there
- If the cursor is visible but laggy or the screen flickers whenever it's not on top of the console window then the game works correctly
- You can test fps in-game if you use `fps_max 999` and `map de_dust2` in the console

Szwagi:
Looked closer to give more details...

[...]

Issue #304, opened 2026-05-13
eDP panel wedges (backlight on, no scanout) after idle-driven DPMS off/on cycle — probabilistic per cycle, regression in 0.11.0


spheenik:
## Summary

After upgrading aquamarine 0.10.0 → 0.11.0, an idle-driven DPMS off/on cycle (e.g. `hypridle` running `hyprctl dispatch "dpms off"` on idle and `dpms on` on resume) intermittently leaves the eDP panel with backlight powered on but no scanout. Hyprland reports `dpmsStatus: 1`, the dispatch returns `ok`, but aquamarine emits no further commits. Once wedged, **only restarting Hyprland recovers** — `hyprctl reload`, additional DPMS cycles, and explicit `hl.monitor({disabled=true})` then `disabled=false` all return `ok` but produce zero new aquamarine output and do not recover the panel.

The bug is **probabilistic — about 15–20% per cycle**. Manual back-to-back `hyprctl dispatch dpms off/on` cycles (without going through the wayland idle protocol path) do **not** reproduce it.

Possibly related to #293 (same backlight-on/no-scanout symptom, same aquamarine 0.11.0 regression, same restart-only recovery), but the trigger is different (idle-driven DPMS, not explicit monitor disable/re-enable).

## System

- **aquamarine:** 0.11.0-2 (broken). 0.10.0-4 was fine (cannot use the package directly with Hyprland 0.55 due to soname bump 9 → 10)
- **Hyprland:** 0.55.0-4
- **Kernel:** 7.0.5.arch1-1
- **Distro:** Arch Linux
- **GPU:** AMD Radeon 780M iGPU (amdgpu)
- **Panel:** eDP-1, Thermotrex TL140ADXP22-0, 2880x1800@60Hz, scale 2.0
- **Idle daemon:** `hypridle` (latest)
- **Reproduced on a second machine:** AMD Ryzen 9 9950X desktop (different amdgpu, no VRR, no eDP) — same wedge symptom following the same idle-driven DPMS pattern

[...]

PR #312, opened 2026-05-28
drm: deterministic page-flip scheduling + coalescing (#304, #306)

spheenik:
## Summary

Replace aquamarine's frame-scheduling state — four loose booleans + a 500 ms staleness timer, mutated from 11 sites — with a per-connector `CFrameScheduler`, deterministic kernel-event-driven invalidation, and a single latest-wins coalescing slot for racing commits. Removes the lossy `"Cannot commit when a page-flip is awaiting"` rejection.

Four independently-readable commits:

1. **`drm: extract CFrameScheduler to own per-connector flip/frame state`** — pure mechanical refactor, zero behaviour change. Wraps `isPageFlipPending`, `pageFlipPendingAtMs`, `frameEventScheduled`, `isFrameRunning` behind named methods on a new connector-owned `CFrameScheduler`. All 11 mutation sites migrated. `needsFrame` stays on `IOutput` (shared by all backends).

2. **`drm: clear in-flight page-flip deterministically on disable and modeset, drop 500ms timer`** — replaces #254's 500 ms `staleFlip` heuristic with two deterministic clears: (a) on **disable**, `applyCommit` clears the flip bookkeeping when the output transitions to disabled (the path #304 needed); (b) on a **blocking modeset**, the kernel guarantees the CRTC reset cancels any in-flight flip (#254's own comment documents this for `ALLOW_MODESET`), so the clear is unconditional — no elapsed-time guess. Deletes `pageFlipPendingAtMs`. Closes the sub-500 ms fast-cycle gap and the non-modeset re-enable gap that the heuristic missed.

3. **`drm: coalesce buffer commits that race an in-flight page-flip`** — when a buffer-only commit arrives while a flip is in flight, stash it in a single latest-wins slot per connector instead of rejecting (lossy — drops the compositor's update, see #306). `handlePF` drains the slot once the pending flip completes, submitting **before** `frame.emit` so the compositor's next render coalesces instead of racing. Buffer lock/release mirrors the existing `last/front/back` discipline; the stash is released on disable / modeset / dead-session.

4. **`drm: pair isFrameRunning set/clear with RAII; settle phase before emit`** — `handlePF`'s manual `setFrameRunning(true/false)` pair was both gated on `session && enabledState` evaluated at different times around event emission; a reentrant state change between them could strand the flag. Replace with a `CFrameRunningGuard` RAII object whose destructor runs on every exit path.

## Issues addressed

- **#304** (idle-DPMS panel wedge) — resolved by commits 1 + 2 (both case-3 disable and case-2 modeset clears are deterministic; no timer).
- **#306** (Source 1 cursor freeze) — the lossy reject is gone; the racing cursor commit is coalesced rather than dropped. The freeze becomes impossible as a *consequence* of no longer dropping the commit. (Cursor-plane reroute — also #306-adjacent — is deliberately out of scope here; happy to follow up.)
- **#254** (the merged 500 ms heuristic) — superseded by commit 2's deterministic clears.

## Not addressed

- **#308** (i915 lid-resume). Its root cause is unconfirmed and the wedge appears to be a *fresh* post-recovery flip whose event is silently dropped — none of the deterministic transition-driven clears here catch that case. I don't have i915 hardware to verify a fix; happy to coordinate with the reporter on a discriminating `AQ_TRACE` if useful.

- **Adjacent observation noticed during testing (not fixed here):** the hardware cursor stays hidden for several frames after a DPMS off→on cycle until either its shape changes or it crosses to another output (which triggers a cursor-plane commit). This looks like cursor-plane state being lost across modeset and not automatically re-armed — orthogonal to the lossy reject this PR fixes. Happy to follow up with a separate PR that primes the cursor plane after modeset, if this lands.

## Testing

Verified locally on amdgpu (Ryzen 9 9950X desktop, multi-monitor) across 30+ cursor-warp + manual-dpms cycles and 25+ wayland-idle DPMS cycles (hypridle 5 s timeout). Zero wedges, zero `Cannot commit when a page-flip is awaiting`, zero `drain of coalesced commit failed`, zero scheduler errors. Each cycle's screen recovery observed visually.

One real S3 suspend/resume cycle was also verified: `restoreAfterVT` fires as expected, no flags are stranded at restore time, no wedge, screen comes back cleanly. Worth disclosing that **one transient `atomic drm request: failed: Device or resource busy` `ERR` line may appear on resume** — the very next buffer commit reaches the kernel during the brief CRTC-settling window after the resume modeset. The existing `CDRMOutput::commitState` retry-as-modeset fallback (`if (!ok && !data.modeset && ...) { data.modeset=true; data.blocking=true; ... }`) re-issues it as a blocking modeset, waits for the kernel to settle, and succeeds — the compositor never sees the failure. The 500 ms `staleFlip` heuristic previously masked this by rejecting commits at the userspace guard before they reached the kernel; with the heuristic gone, the kernel-level transient is what surfaces in the log instead. Happy to silence the upstream ERR log in a follow-up if desired.

### DIFF
--- a/include/aquamarine/backend/FrameScheduler.hpp
+++ b/include/aquamarine/backend/FrameScheduler.hpp
@@ -13,51 +13,31 @@ namespace Aquamarine {
         // a frame was submitted to the host with a completion event requested.
         // DRM: page-flip submitted with PAGE_FLIP_EVENT. Wayland: sendCommit followed
         // by a wl_surface.frame whose done is wired to onFrameComplete.
-        void onFrameSubmitted() {
-            m_pending = true;
-        }
+        void onFrameSubmitted();
 
         // the host delivered the frame completion event.
         // DRM: kernel page-flip handler. Wayland: wl_callback.done.
         // State-only — clears m_pending. The backend fires frameReady explicitly
         // at the right point in its handler (preserving the present-before-frame
         // ordering on DRM).
-        void onFrameComplete() {
-            m_pending = false;
-        }
+        void onFrameComplete();
 
         // the in-flight frame (if any) is void and no completion event will arrive;
         // reset the frame bookkeeping.
-        void invalidate() {
-            m_pending        = false;
-            m_frameRunning   = false;
-            m_frameScheduled = false;
-        }
+        void invalidate();
 
         // is a frame in flight awaiting completion?
-        bool frameInFlight() const {
-            return m_pending;
-        }
+        bool frameInFlight() const;
 
         // may a new frame be scheduled? (does not consider output enabled state)
-        bool canSchedule() const {
-            return !m_pending && !m_frameRunning && !m_frameScheduled;
-        }
+        bool canSchedule() const;
 
         // State accessors. setFrameScheduled / frameScheduled() drive the idle-frame
         // loop; setFrameRunning / frameRunning() guard event emission on completion.
-        bool frameScheduled() const {
-            return m_frameScheduled;
-        }
-        void setFrameScheduled(bool v) {
-            m_frameScheduled = v;
-        }
-        bool frameRunning() const {
-            return m_frameRunning;
-        }
-        void setFrameRunning(bool v) {
-            m_frameRunning = v;
-        }
+        bool frameScheduled() const;
+        void setFrameScheduled(bool v);
+        bool frameRunning() const;
+        void setFrameRunning(bool v);
 
         // Fires from onFrameComplete. The output wires this to events.frame.emit.
         Hyprutils::Signal::CSignalT<> frameReady;
@@ -72,12 +52,8 @@ namespace Aquamarine {
     // reentrant emit handler cannot strand it.
     class CFrameRunningGuard {
       public:
-        explicit CFrameRunningGuard(CFrameScheduler& s) : m_s(s) {
-            m_s.setFrameRunning(true);
-        }
-        ~CFrameRunningGuard() {
-            m_s.setFrameRunning(false);
-        }
+        explicit CFrameRunningGuard(CFrameScheduler& s);
+        ~CFrameRunningGuard();
         CFrameRunningGuard(const CFrameRunningGuard&)            = delete;
         CFrameRunningGuard& operator=(const CFrameRunningGuard&) = delete;
 
--- a/include/aquamarine/backend/FrameScheduler.hpp
+++ b/include/aquamarine/backend/FrameScheduler.hpp
@@ -1,33 +1,41 @@
 #pragma once
 
+#include <hyprutils/signal/Signal.hpp>
+
 namespace Aquamarine {
-    // Per-connector page-flip / frame scheduling state for the DRM backend.
+    // Per-output frame scheduling state, shared by the DRM and Wayland backends.
     //
-    // Wraps the flip-flags that previously lived directly on SDRMConnector
-    // (isPageFlipPending, frameEventScheduled, isFrameRunning) behind named methods,
-    // so a single owner holds the page-flip/frame lifecycle.
+    // Both backends produce an end-of-frame event — DRM from the kernel page-flip
+    // completion, Wayland from wl_callback.done on a wl_surface.frame. This class
+    // models that lifecycle uniformly and fires frameReady on completion.
     class CFrameScheduler {
       public:
-        // a page-flip was submitted to the kernel with a completion event requested.
-        void onFlipSubmitted() {
+        // a frame was submitted to the host with a completion event requested.
+        // DRM: page-flip submitted with PAGE_FLIP_EVENT. Wayland: sendCommit followed
+        // by a wl_surface.frame whose done is wired to onFrameComplete.
+        void onFrameSubmitted() {
             m_pending = true;
         }
 
-        // the kernel delivered the page-flip completion event.
-        void onFlipComplete() {
+        // the host delivered the frame completion event.
+        // DRM: kernel page-flip handler. Wayland: wl_callback.done.
+        // State-only — clears m_pending. The backend fires frameReady explicitly
+        // at the right point in its handler (preserving the present-before-frame
+        // ordering on DRM).
+        void onFrameComplete() {
             m_pending = false;
         }
 
-        // the in-flight flip (if any) is void and no completion event will arrive;
-        // reset the flip/frame bookkeeping.
+        // the in-flight frame (if any) is void and no completion event will arrive;
+        // reset the frame bookkeeping.
         void invalidate() {
             m_pending        = false;
             m_frameRunning   = false;
             m_frameScheduled = false;
         }
 
-        // is a page-flip in flight on this connector's CRTC?
-        bool flipInFlight() const {
+        // is a frame in flight awaiting completion?
+        bool frameInFlight() const {
             return m_pending;
         }
 
@@ -37,7 +45,7 @@ namespace Aquamarine {
         }
 
         // State accessors. setFrameScheduled / frameScheduled() drive the idle-frame
-        // loop; setFrameRunning / frameRunning() guard event emission in handlePF.
+        // loop; setFrameRunning / frameRunning() guard event emission on completion.
         bool frameScheduled() const {
             return m_frameScheduled;
         }
@@ -50,6 +58,9 @@ namespace Aquamarine {
         void setFrameRunning(bool v) {
             m_frameRunning = v;
         }
+
+        // Fires from onFrameComplete. The output wires this to events.frame.emit.
+        Hyprutils::Signal::CSignalT<> frameReady;
 
       private:
         bool m_pending        = false;
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -408,11 +408,9 @@ void Aquamarine::CDRMBackend::restoreAfterVT() {
     // restore, but isPageFlipPending is already false so the = false
     // assignment is a harmless no-op.
     for (auto const& c : connectors) {
-        if (c->isPageFlipPending || c->isFrameRunning) {
+        if (c->sched.flipInFlight() || c->sched.frameRunning()) {
             backend->log(AQ_LOG_DEBUG, std::format("drm: Clearing stale page-flip state for {}", c->szName));
-            c->isPageFlipPending   = false;
-            c->isFrameRunning      = false;
-            c->frameEventScheduled = false;
+            c->sched.invalidate();
         }
     }
 
@@ -1142,7 +1140,7 @@ static void handlePF(int fd, unsigned seq, unsigned tv_sec, unsigned tv_usec, un
     if (!pageFlip || !pageFlip->connector)
         return;
 
-    pageFlip->connector->isPageFlipPending = false;
+    pageFlip->connector->sched.onFlipComplete();
 
     const auto& BACKEND = pageFlip->connector->backend;
 
@@ -1154,7 +1152,7 @@ static void handlePF(int fd, unsigned seq, unsigned tv_sec, unsigned tv_usec, un
     }
 
     if (BACKEND->sessionActive() && pageFlip->connector->output->enabledState)
-        pageFlip->connector->isFrameRunning = true;
+        pageFlip->connector->sched.setFrameRunning(true);
 
     pageFlip->connector->onPresent();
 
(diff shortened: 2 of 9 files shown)


vaxerski:
yeah, was long overdue. The next step would be fixing scheduling on wayland. Considering both those backends produce frame-ready events, could we use the scheduler for both?

Install an emittable event in the scheduler, make pfHandler (drm) or frameCb (wayland) trigger it. Functionally, they should be the same.

spheenik:
Pushed two more commits lifting the scheduler to the Wayland backend, per the discussion above.

**`backend: lift CFrameScheduler to a shared per-output scheduler`** — moves `CFrameScheduler` out of `backend/drm/` and generalises the vocabulary (`onFlipSubmitted`/`onFlipComplete`/`flipInFlight` → `onFrameSubmitted`/`onFrameComplete`/`frameInFlight`). Adds the emittable `frameReady` signal you suggested; each output wires it to its public `events.frame` in its constructor. `onFrameComplete` is state-only, so the backend fires `frameReady` explicitly at the right point and DRM keeps its present-before-frame ordering in `handlePF`. No behaviour change on the DRM path — it just routes `events.frame` through the signal (one synchronous hop); the coalesce slot stays DRM-specific.

**`wayland: drive frame scheduling through CFrameScheduler`** — replaces the Wayland backend's ad-hoc loop (three bools + the `onFrameDone` band-aid from 46ce868) with the shared scheduler. `onFrameDone` now mirrors `handlePF`: `onFrameComplete()` → `CFrameRunningGuard` → present, then `frameReady`. The loop self-limits the same way DRM's does — if the consumer doesn't commit (its `needsFrame` clears in `renderMonitor`), no new `wl_surface.frame` is registered and the loop stops until the next `scheduleFrame`.

The band-aid was forcing `needsFrame = true` on every `wl_callback.done`, so the backend free-ran at the host refresh rate regardless of damage. Removing it drops an idle nested Hyprland from ~60 `scheduleFrame`/s to **0**, while focused input still renders immediately.

A few Wayland-specific points that each cause a hang if missed, for reviewer attention:
- the `wl_surface.frame` request must precede `sendCommit()` (it's pending surface state, applied on the next commit);
- it's gated on `!frameInFlight()` so a rapid cursor+buffer double-commit shares the in-flight callback rather than orphaning the first;
- the display is flushed right after `sendCommit()`, since `commit()` runs from the consumer's render path outside `dispatchEvents()` — without it the callback request sits unflushed and the loop deadlocks;
- the frame-emit idle goes through the core backend's `addIdleEvent` (pumped every loop iteration), not the backend-local `idleCallbacks` (drained only when the wl fd is readable, which stalls after a focus change with no host traffic). `~CWaylandOutput` removes it to avoid firing on a freed output.

Tested via a nested Hyprland (Wayland backend): idle is silent (0 `scheduleFrame`), focused typing/cursor render immediately and survive unfocus→refocus, and client/output teardown is clean. DRM session unaffected on a real relog.

vaxerski:
hey tom @gulafaran can you test this?

src/backend/drm/DRM.cpp
```diff
@@ -1144,10 +1143,24 @@ static void handlePF(int fd, unsigned seq, unsigned tv_sec, unsigned tv_usec, un
...
+            SDRMConnectorCommitData draining = std::move(*pageFlip->connector->nextCommit);
+            pageFlip->connector->nextCommit.reset();
+            if (!pageFlip->connector->commitState(draining))
+                BACKEND->log(AQ_LOG_ERROR, std::format("drm: drain of coalesced commit failed for {}", pageFlip->connector->szName));
+        } else {
+            // Session inactive, output disabled, or a reentrant present.emit
+            // handler already submitted a newer flip — stash is obsolete, drop it.
+            pageFlip->connector->releaseStashedCommit();
+        }
     }
```
gulafaran: what is this supposed todo? if `pageFlip->connector->nextCommit` commit things but if not? it calls "releaseStashedCommit();" 

```
void Aquamarine::SDRMConnector::releaseStashedCommit() {
    if (!nextCommit)              <----------------------------------- this just returns?
        return;
    if (nextCommit->mainFB) {
        nextCommit->mainFB->buffer->lockedByBackend = false;
        nextCommit->mainFB->buffer->events.backendRelease.emit();
    }
    if (crtc && crtc->cursor && nextCommit->cursorFB) {
        nextCommit->cursorFB->buffer->lockedByBackend = false;
        nextCommit->cursorFB->buffer->events.backendRelease.emit();
    }
    nextCommit.reset();
}

```

gulafaran (changes requested):

src/backend/drm/DRM.cpp
spheenik: Good catch — the structure was confusing even though the behaviour was fine, so I've reworked it (pushed in 2840aa3).

To answer the literal question first: in that `else` branch `nextCommit` is always set — the whole block was under `if (nextCommit)` and only the drain arm `.reset()`s it — so `releaseStashedCommit()` did do the release, it didn't early-return. The `if (!nextCommit) return;` inside it isn't for that call site; it guards the three *unconditional* callers (`applyCommit` on disable, modeset, dead-session teardown), where there may be nothing stashed. And the branch itself isn't dead: `onFrameComplete()` clears `frameInFlight` at the top of `handlePF`, then `present.emit` can reentrantly drive a commit that — seeing no flip in flight — submits a new flip directly and re-arms `frameInFlight`; by the time we reach the drain the old stash is stale and submitting it would `-EBUSY`, so it's dropped.

That said, you were right that calling a re-guarded helper after we've already proven the stash is present reads badly. So rather than defend it:

- the drain logic now lives in a named `SDRMConnector::drainStashedCommit()` with a named `canDrain` condition (handlePF just calls it);
- the buffer-release body is split into an unguarded `releaseCommitBuffers(commit)` core;
- `releaseStashedCommit()` keeps its guard purely for its three unconditional callers, and the drain drop-path calls the unguarded core directly.

Net: the release body exists in one place and no path re-checks `nextCommit`. No behaviour change.

Skidamek:
hi, can you rebase?

spheenik:
Rebased onto current `main` (0.12.1) and reconciled with #325, which landed in the same frame-scheduling area while this was open. Force-pushed; commit hashes changed.

### Reconciliation with #325
#325's "only one `.frame` per pageflip" guarantees now live inside `CFrameScheduler` instead of the loose booleans:
- the `handlePF` *"an idle frame is already queued — defer to it"* guard → `if (sched.frameScheduled()) return;` before `frameReady.emit()` (the RAII guard clears `frameRunning` on the early return);
- the `frameIdle` *"a newer flip got armed — drop the redundant idle"* tail → translated to the scheduler accessors;
- the new `pendingIdleFrame()` → `sched.frameScheduled()` on DRM/Wayland (Headless keeps its own bool).

Worth flagging that these guards are **not** redundant in the scheduler model — I checked carefully. The coexistence state (an idle frame queued *and* a flip in flight) is reachable when `scheduleFrame()` runs before the commit that arms the flip; without the `handlePF` guard the pf handler and the queued idle both emit. The pre-#325 version of this branch actually had that exact latent double-emit, so merging #325 in closed a real gap rather than just resolving a textual conflict.

### Review feedback
Addressed @gulafaran's note on the coalesce drain: the inline drain-or-drop is now a named `SDRMConnector::drainStashedCommit()`, and the buffer-release body is split into an unguarded `releaseCommitBuffers()` core, so no path re-checks `nextCommit` after the guard already established it.

### Testing (real hardware)
amdgpu, Ryzen 9 9950X desktop, multi-monitor, Hyprland 0.55.4 against this build (aquamarine 0.12.1, soname 11). 12+ manual DPMS off/on + cursor-warp cycles plus repeated input-driven wakes: zero `Cannot commit when a page-flip is awaiting`, zero `drain of coalesced commit failed`, zero scheduler errors; both outputs recover every cycle.

### Adjacent observation (not addressed here — likely in the "fix scheduling on wayland" space)
After `dpms on`, Hyprland reports the output as on, but aquamarine's connector `enabledState` stays false until the next commit — and with no damage there's no commit, so the panel stays physically off until input generates damage. In other words no frame is primed on (re-)enable. This is pre-existing (reproduces independent of this PR), but it may be worth priming a frame when an output transitions to enabled. Happy to look into it as a follow-up.

gulafaran:
Got no other questions, seems to run fine here

Honkazel:
Running this one for days. Seems to work okay, can't repro #308 on my comet lake laptop
@johnkferguson Could you test it too? Since you're the one, who reported this issue.

johnkferguson:
@Honkazel I [posted an update](https://github.com/hyprwm/aquamarine/issues/308#issuecomment-4794379017) in #308.

The short of it is that my issue was a hypridle config issue that occured when I updated to Hyprland lua config, not actually aquamarine related.

So this PR shouldn't actually change anything for me. I haven't tested against this branch as it likely shouldn't be needed at this point, but if you want me to run it to check for potential regressions in my setup, let me know and I can do so.

include/aquamarine/backend/FrameScheduler.hpp
```diff
@@ -0,0 +1,87 @@
...
+    //
+    // Both backends produce an end-of-frame event — DRM from the kernel page-flip
+    // completion, Wayland from wl_callback.done on a wl_surface.frame. This class
+    // models that lifecycle uniformly and fires frameReady on completion.
+    class CFrameScheduler {
+      public:
+        // a frame was submitted to the host with a completion event requested.
+        // DRM: page-flip submitted with PAGE_FLIP_EVENT. Wayland: sendCommit followed
+        // by a wl_surface.frame whose done is wired to onFrameComplete.
+        void onFrameSubmitted() {
```
vaxerski: all these functions should have bodies in a cpp file not in the public header.

spheenik:
Moved all `CFrameScheduler` / `CFrameRunningGuard` method bodies out of the public header into a new `src/backend/FrameScheduler.cpp`; the header now carries declarations only. No behaviour change. (`CONFIGURE_DEPENDS` picks the new `.cpp` up automatically — no CMake edit needed.)

Also rebased onto current `main` while here — the two commits that landed since (`c49d9c4`, `6d6e238`) are one-line `!!` bool-cast fixes in `Backend.cpp`/`DRM.cpp`, both clear of this PR's changes, so the rebase was clean.

Re-tested on the Wayland backend (nested Hyprland against the rebuilt lib): output and client come up, idle stays silent (loop parked in `epoll_wait`, 0 `scheduleFrame`), input-driven renders are immediate, and teardown is clean — no commit/scheduler errors across the session.

src/backend/drm/DRM.cpp
```diff
@@ -1170,14 +1169,20 @@ static void handlePF(int fd, unsigned seq, unsigned tv_sec, unsigned tv_usec, un
         .flags     = flags,
     });
 
+    // Drain any buffer-only commit stashed while this flip was in flight, BEFORE
+    // frame.emit — frame.emit triggers the compositor's next commit, which would
+    // otherwise race the drain and hit -EBUSY.
+    pageFlip->connector->drainStashedCommit();
```
vaxerski: I dont like this, we're delaying commits potentially forever. Why not drain the state, but only commit if frameReady.emit() did not result in a commit?

spheenik: Good catch — reordered so the emit runs first and the drain after (`e684a79`). Draining before the emit re-armed the flip, so every fresh commit landed while it was busy and got stashed: the rendered frame stayed one flip behind for as long as the coalesce path was engaged, which is the "delaying forever" you're pointing at.

Now `frameReady.emit()` runs while the flip is free, so the compositor's fresh commit submits directly; `drainStashedCommit()` then runs and its existing `!frameInFlight()` guard drops the stash as stale — or submits it only when nothing fresher rendered. No `-EBUSY` risk, since the guard never commits while a flip is in flight, and the pipeline falls straight back to the zero-latency direct-submit path.

Tested on amdgpu real hardware (Ryzen 9950X, DP-1 @100Hz + HDMI-A-1 @60Hz): steady-state desktop use, cursor-commit-vs-flip coalesce churn, and 3 DPMS off/on cycles. Both outputs re-enable cleanly every cycle; zero `Cannot commit when a page-flip is awaiting`, zero `drain of coalesced commit failed`, zero EBUSY.

vaxerski (approved):

gulafaran (approved):

PR #323, opened 2026-06-13
drm: fix resource leaks

ghost:
Fix resource leaks and missing cleanup in the DRM backend.

- Free the disconnected connectors returned by `drmModeGetConnector`
- Free the string returned by `drmGetDeviceNameFromFd2`
- Free the model and serial strings returned by libdisplay-info
- Clean up DRM resources on `initResources` failure paths
- Release the udev enumeration when no GPUs are found
- Free the connector data when connector initialization fails
- Close the DRM fd when `drmDropMaster` fails

The startup leaks were identified with Valgrind.

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -19,7 +19,7 @@
 #include <unordered_set>
 #include <sys/mman.h>
 #include <fcntl.h>
-
+#include <unistd.h>
 extern "C" {
 #include <libseat.h>
 #include <libudev.h>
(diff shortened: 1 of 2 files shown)


src/backend/drm/DRM.cpp
```diff
@@ -594,13 +595,17 @@ bool Aquamarine::CDRMBackend::initResources() {
...
+        drmModeFreeResources(resources);
+        crtcs.clear();
         return false;
     }
 
     // initialize planes
     auto planeResources = drmModeGetPlaneResources(gpu->fd);
     if (!planeResources) {
         backend->log(AQ_LOG_ERROR, std::format("drm: drmModeGetPlaneResources failed"));
+        drmModeFreeResources(resources);
```
vaxerski: scopeguard perhaps?

ghost: i found one more leak, i seriusly think the tests should run with valgring memcheck or something as default

vaxerski: ts has no tests we should write some

vaxerski (approved):

PR #327, opened 2026-06-28
backend: enable libinput plugins if available

mechakotik:


vaxerski:
where are those loaded from? What do other compositors do? Security-wise.

mechakotik:
KWin loads libinput plugins in the same way, it is enabled by default, but can be disabled with env var:

```cpp
#if HAVE_LIBINPUT_PLUGINS
const bool wantsPlugins = !environmentVariableBoolValue("KWIN_LIBINPUT_NO_PLUGINS").value_or(false);
if (wantsPlugins) {
    libinput_plugin_system_append_default_paths(m_libinput);
    libinput_plugin_system_load_plugins(m_libinput, LIBINPUT_PLUGIN_SYSTEM_FLAG_NONE);
}
#endif
```

Load paths are defined by [`libinput_plugin_system_append_default_paths`](https://wayland.freedesktop.org/libinput/doc/latest/api/group__base.html#ga85b12a6119c930324dfc0b9986977819) and are typically:

```
/etc/libinput/plugins/
/usr/lib{64}/libinput/plugins/
```

Currently there are only root-controlled paths, which is pretty safe, but this set of paths may change in future, and libinput docs give no guarantee they won't add user-controlled paths. So if we want to narrow down an attack surface, maybe we should not append default paths, but some hardcoded list of paths like this:

```cpp
libinput_plugin_system_append_path(session->libinputHandle, "/etc/libinput/plugins");
libinput_plugin_system_append_path(session->libinputHandle, "/usr/lib64/libinput/plugins");
libinput_plugin_system_append_path(session->libinputHandle, "/usr/lib/libinput/plugins");
```

Though I don't think they will make such an insecure change.

vaxerski:
that's fine then. I'd still add an env just like kwin to not load.

src/backend/Session.cpp
```diff
@@ -344,14 +345,29 @@ SP<CSession> Aquamarine::CSession::attempt(Hyprutils::Memory::CSharedPointer<CBa
         return nullptr;
     }
 
+    libinput_log_set_handler(session->libinputHandle, ::libinputLog);
+    libinput_log_set_priority(session->libinputHandle, LIBINPUT_LOG_PRIORITY_DEBUG);
+
+#ifdef AQUAMARINE_HAS_LIBINPUT_PLUGINS
+    if (envEnabled("AQ_LIBINPUT_NO_PLUGINS")) {
```
vaxerski: no {}

vaxerski (commented):
other than that lgtm

vaxerski (approved):
tnx

PR #340, opened 2026-07-23
drm: dont set HW_CLOCK on nvidia reporting seq 0

gulafaran:
dont set the the clock as HW_CLOCK on nvidia when sequence is 0, its a plain ktime_get() software time whenever the event was called. because nvidia doesnt register vblank counters without manually setting nvidia_drm vblank=1


vaxerski (approved):
surprisingly polite comment, I would've put some fucks in there

PR #350, opened 2026-08-04
drm: keep scanning extension blocks after the first CTA

willmtemple:
Changes EDID scanning to visit all extension blocks, and picks HDR static metadata and colorimetry from the first CTA extension blocks that contain each one to be defensive against potentially bad EDIDs.

This is relatively uncommon as far as I know for monitors but can help with TVs that have multiple CTA extension blocks (they sometimes have lots of AV timing data and can't fit it in a single block) and HDR/colorimetry might not be in the first one. The spec only requires that HDR static metadata and colorimetry be singletons within the whole EDID payload, not that they be located on the same CTA extension block or that there only be one CTA extension block.

As a side effect, users whose monitors have a DisplayID extension block after a CTA extension block will start seeing the FIXME for DisplayID blocks.

### DIFF
--- a/src/backend/drm/DRM.cpp
+++ b/src/backend/drm/DRM.cpp
@@ -1642,6 +1642,10 @@ IOutput::SParsedEDID Aquamarine::SDRMConnector::parseEDID(std::vector<uint8_t> d
 
     auto exts = di_edid_get_extensions(edid);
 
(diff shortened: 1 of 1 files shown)


vaxerski:
@codex review please

vaxerski (approved):
lgtm :fireemoji:

PR #356, opened 2026-08-15
swapchain: preserve multigpu flag in resize()

Kevi14:
fullReconfigure() passes .multigpu when acquiring buffers, resize() did not, so any reallocation that only changed the swapchain length silently dropped it.

Issue caught when using dual monitor setup with multi gpu: Intel integrated + Nvidia RTX.


vaxerski (approved):
thanks!

PR #366, opened 2026-08-25
drm: reserve CRTCs used by redundant tiles

s0up4200:
Aquamarine excluded redundant tile connectors from its occupied-CRTC scan even when AMDGPU kept them connected to active CRTCs. With two Apple Studio Displays, it assigned `DP-4`'s CRTC to `DP-1`, causing atomic commits to fail with `EINVAL` and leaving `DP-1` black at `0x0`. The patch counts cached CRTCs for all connected connectors and leaves physical connector status to the DRM scan.

Built and tested against Aquamarine 0.14.0 with Hyprland 0.56.2 on a Radeon 780M. Both displays started and reconnected  without `EINVAL`

One screen is connected via Thunderbolt, and the other with one of these: https://www.amazon.com/gp/product/B0BNX7MS6N recommended in the Omarchy docs: https://learn.omacom.io/2/the-omarchy-manual/86/monitors#apple-displays

Using a Beelink SER8 machine.

AI disclosure: Codex using GPT-5.6 Sol with xhigh reasoning wrote this change under my supervision.


src/backend/drm/DRM.cpp
```diff
@@ -1877,7 +1878,6 @@ void Aquamarine::SDRMConnector::disconnect() {
 
     invalidateFrame();
 
-    status = DRM_MODE_DISCONNECTED;
```
vaxerski, line 2136: why?

s0up4200: because removing the status assignment in disconnect() prevents recheckCRTCs() from clearing the cached CRTC of physically connected redundant tiles

vaxerski: then that should be fixed instead of leaving status of CONNECTED on connectors when disconnect() was called.

s0up4200: would this be better? `disconnect()` still sets the status, while `recheckCRTCs()` keeps the cached CRTC for redundant tiles.

vaxerski (approved):
this looks good

